PR #7, opened 2026-04-27
Nightly Bug Sweep — web — 2026-04-27 — Batch 01

jack-opsapp:
## Summary

Nightly automated bug sweep for OPS-Web. Batch 01 of 02 (10 bugs).

- 5 bugs fixed in code (4 distinct commits — bugs 94977186 / 8665d78e share the keyboard-shortcut fix)
- 5 bugs escalated for human review

## Fixed

- [x] **bug-916ca622** [COMPLEX] — functional/high — Typing in project Notes triggered global shortcuts. Hardened the global `KeyboardShortcuts` listener to bail when EITHER `e.target` OR `document.activeElement` is editable. Added a tag-name fallback for portals and a contenteditable ancestor walk. `src/components/ops/keyboard-shortcuts.tsx`.
- [x] **bug-02c5d49c** — functional/medium — Project creation blocked without guidance. Added an invalid-submit handler that surfaces a top-level error banner and scrolls/focuses the first invalid field; marked the title field as Required (helper text + aria-required). `src/app/(dashboard)/projects/new/page.tsx`, dictionaries.
- [x] **bug-94977186** [COMPLEX] — functional/high — Project creation form inputs inaccessible. Likely same root cause as 916ca622; covered by the keyboard-shortcuts hardening. Re-open with a DOM snapshot if the symptom persists.
- [x] **bug-8665d78e** [COMPLEX] — functional/high — Client creation form inputs inaccessible. Likely same root cause as 916ca622; covered by the keyboard-shortcuts hardening. Re-open with a DOM snapshot if the symptom persists.
- [x] **bug-5b32195c** [COMPLEX] — ui/high — Mobile dashboard sidebar overlapping at 390px. Paired the existing `-translate-x-full` with `invisible pointer-events-none` on mobile-closed and re-asserted visibility/pointer-events at md+. Added `aria-hidden` for assistive tech. `src/components/layouts/sidebar.tsx`.
- [x] **bug-d647b755** — ui/medium — Timeline overlapping events stacking unreadably. Added a sweep-line lane assignment per row; overlapping events get distinct lanes (≥24px each, 4px gaps) and the row height grows to fit. `src/app/(dashboard)/calendar/_components/timeline/timeline-grid.tsx`, `timeline-row.tsx`, `timeline-task-block.tsx`.

## Escalated (requires human review)

- **bug-7e5806cf** [COMPLEX] — Hotmail login fails for Dana Hartford. External Microsoft account state, not an OPS-Web code change.
- **bug-7aa403c7** [COMPLEX] — No OPS / Maverick emails in test client inboxes. Missing test seed data; staging needs portal invitation / estimate / invoice generation, not a code fix.
- **bug-83cba0f6** [COMPLEX] — Tutorial-Interactive freezes on step 2. Lives on try.opsapp.co (try-ops repo), not ops-web. Out of scope for this batch.
- **bug-a30886bd** — Calendar drag-and-drop non-functional. Inspected month + timeline DnD setups (dnd-kit useDraggable, PointerSensor distance:8) and surrounding pointer-events / z-index stack. Code path looks correct; without a browser repro I can't determine whether activation distance is being missed, a parent overlay is intercepting pointerdown, or listeners are detached. Needs a human with React DevTools / dnd-kit debug.

## Build

- `npm run type-check` — clean.
- `npm run build` — fails with `supabaseUrl is required` during page-data collection. This is a pre-existing environment issue on `main` (verified by checking out `main` and re-running build); not caused by changes in this batch.

## Test plan

- [ ] Open `/projects/<id>?tab=notes`, focus the Notes textarea, type "1234567890". Should enter literally; should NOT navigate to `/dashboard`, `/projects`, etc.
- [ ] Open `/projects/new`, leave title empty, click Create Project. Should show top-level red banner "Please fix the highlighted fields..." and scroll/focus the title field.
- [ ] Open dashboard at 390px viewport. Sidebar should be fully hidden until hamburger is tapped; widgets should fill full width.
- [ ] Open calendar timeline view with two overlapping events on the same team member's row. Both should render in distinct vertical lanes; both labels readable; both clickable.

### DIFF
--- a/src/components/layouts/sidebar.tsx
+++ b/src/components/layouts/sidebar.tsx
@@ -274,14 +274,23 @@ export function Sidebar() {
       <aside
         onMouseEnter={() => { if (!isMobileView) setHoverExpanded(true); }}
         onMouseLeave={() => { if (!isMobileView) setHoverExpanded(false); }}
+        aria-hidden={isMobileView && !isMobileOpen ? "true" : undefined}
         className={cn(
           "fixed left-0 top-0 h-screen z-[45]",
           "border-r border-[rgba(255,255,255,0.06)]",
           "flex flex-col transition-all duration-200 ease-out",
           effectiveCollapsed ? "w-[72px]" : "w-[256px]",
-          // Mobile: off-screen by default, slide in when open
+          // Mobile: off-screen by default, slide in when open. We pair the
+          // translate with invisible/pointer-events-none so the drawer is
+          // genuinely removed from interaction even if the transform fails
+          // to apply (some legacy mobile browsers / WebView rendering paths
+          // dropped the sidebar back into layout, which left it overlapping
+          // dashboard widgets at 390px).
           isMobileOpen ? "translate-x-0" : "-translate-x-full",
-          "md:translate-x-0"
+          isMobileOpen ? "visible pointer-events-auto" : "invisible pointer-events-none",
+          // Restore visibility/interaction at md+ regardless of mobile-open
+          // state so the desktop hover-rail keeps working.
+          "md:translate-x-0 md:visible md:pointer-events-auto"
         )}
         style={{
           background: "var(--surface-glass-dense)",
--- a/src/components/ops/keyboard-shortcuts.tsx
+++ b/src/components/ops/keyboard-shortcuts.tsx
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
   const router = useRouter();
 
   useEffect(() => {
+    function isEditableElement(node: Element | null | undefined): boolean {
+      if (!node) return false;
+      if (node instanceof HTMLInputElement || node instanceof HTMLTextAreaElement) {
+        return true;
+      }
+      if (node instanceof HTMLElement && node.isContentEditable) {
+        return true;
+      }
+      // Tag-name fallback for cases where instanceof fails (e.g. cross-realm
+      // events dispatched from portals, iframes, or libraries that re-host the
+      // node). Also covers select elements and standard form controls.
+      const tag = node.tagName?.toUpperCase();
+      if (tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT") {
+        return true;
+      }
+      // Walk up looking for a contenteditable ancestor.
+      if (node instanceof HTMLElement && node.closest("[contenteditable='true'], [contenteditable='']")) {
+        return true;
+      }
+      return false;
+    }
+
     function handleKeyDown(e: KeyboardEvent) {
-      // Don't capture if user is typing in an input
-      const target = e.target as HTMLElement;
-      if (
-        target instanceof HTMLInputElement ||
-        target instanceof HTMLTextAreaElement ||
-        target.isContentEditable ||
-        target.closest("[contenteditable]")
-      ) {
+      // Don't capture if user is typing in an input — check both the event
+      // target AND the document's active element. The active-element fallback
+      // catches cases where keystrokes are dispatched after a focus/blur race
+      // or where the event target is the body/window rather than the field
+      // itself (observed when a controlled textarea re-renders mid-keystroke).
+      const target = e.target as Element | null;
+      if (isEditableElement(target) || isEditableElement(document.activeElement)) {
         return;
       }
 
--- a/src/app/(dashboard)/calendar/_components/timeline/timeline-grid.tsx
+++ b/src/app/(dashboard)/calendar/_components/timeline/timeline-grid.tsx
@@ -24,7 +24,71 @@ import {
   TIMELINE_DAYS_SHOWN,
   TIMELINE_DAY_MIN_WIDTH,
   TIMELINE_GUTTER_WIDTH,
+  TIMELINE_ROW_HEIGHT,
 } from "@/lib/utils/timeline-constants";
+
+// ─── Lane assignment ────────────────────────────────────────────────────────
+
+interface LaneAssignment {
+  /** eventId → lane index */
+  lanes: Map<string, number>;
+  /** maximum number of overlapping lanes used */
+  laneCount: number;
+}
+
+/**
+ * Sweep-line lane assignment: events sorted by start date are placed into the
+ * lowest-numbered lane whose previous event has already ended. Two events
+ * count as overlapping when their inclusive [start, end] ranges intersect.
+ */
+function assignLanes(events: InternalCalendarEvent[]): LaneAssignment {
+  if (events.length === 0) return { lanes: new Map(), laneCount: 1 };
+
+  const sorted = [...events].sort(
+    (a, b) => a.startDate.getTime() - b.startDate.getTime()
+  );
+  // laneEndTimes[i] = end timestamp of the latest event currently in lane i
+  const laneEndTimes: number[] = [];
+  const lanes = new Map<string, number>();
+
+  for (const event of sorted) {
+    const startMs = event.startDate.getTime();
+    const endMs = event.endDate.getTime();
+    let placed = false;
+
+    for (let i = 0; i < laneEndTimes.length; i++) {
+      // Overlap if this event starts on or before the lane's last event ends.
+      // Use strict less-than so an event ending on day N can sit in the same
+      // lane as one starting on day N+1 (back-to-back, not overlapping).
+      if (laneEndTimes[i] < startMs) {
+        lanes.set(event.id, i);
+        laneEndTimes[i] = endMs;
+        placed = true;
+        break;
+      }
+    }
+
+    if (!placed) {
+      lanes.set(event.id, laneEndTimes.length);
+      laneEndTimes.push(endMs);
+    }
+  }
+
+  return { lanes, laneCount: Math.max(laneEndTimes.length, 1) };
+}
+
+/**
+ * Decide row height given a lane count. Each lane is allotted ≥24px so labels
+ * stay readable; rows always honor the base TIMELINE_ROW_HEIGHT minimum.
+ */
+function rowHeightForLanes(laneCount: number): number {
+  const MIN_LANE_HEIGHT = 24;
+  const VERTICAL_PADDING = 16; // 8px top + 8px bottom
+  const LANE_GAP = 4;
+  const computed =
+    VERTICAL_PADDING + laneCount * MIN_LANE_HEIGHT + Math.max(laneCount - 1, 0) * LANE_GAP;
+  return Math.max(TIMELINE_ROW_HEIGHT, computed);
+}
 
 // ─── Unassigned Placeholder ─────────────────────────────────────────────────
 
@@ -59,12 +123,14 @@ function DroppableTimelineRow({
   startDate,
   daysShown,
   isLast,
+  rowHeight,
   children,
 }: {
   teamMember: TeamMember;
   startDate: Date;
   daysShown: number;
   isLast?: boolean;
+  rowHeight?: number;
   children: React.ReactNode;
 }) {
   const rowRef = useRef<HTMLDivElement>(null);
@@ -108,6 +174,7 @@ function DroppableTimelineRow({
         startDate={startDate}
         daysShown={daysShown}
         isLast={isLast}
+        rowHeight={rowHeight}
       >
         {children}
       </TimelineRow>
@@ -305,12 +372,15 @@ export function TimelineGrid({
           {/* Team member rows */}
           {teamMembers.map((member) => {
             const memberEvents = grouped.get(member.id) ?? [];
+            const { lanes, laneCount } = assignLanes(memberEvents);
+            const rowHeight = rowHeightForLanes(laneCount);
             return (
               <DroppableTimelineRow
                 key={member.id}
                 teamMember={member}
                 startDate={startDate}
                 daysShown={daysShown}
+                rowHeight={rowHeight}
               >
                 {memberEvents.map((event) => (
                   <TimelineTaskBlock
@@ -319,6 +389,9 @@ export function TimelineGrid({
                     startDate={startDate}
                     daysShown={daysShown}
                     isSelected={isTaskSelected(event.id)}
+                    laneIndex={lanes.get(event.id) ?? 0}
+                    laneCount={laneCount}
+                    rowHeight={rowHeight}
                     onClick={handleEventClick}
                     onContextMenu={handleContextMenu}
                     onResize={handleResize}
@@ -332,12 +405,15 @@ export function TimelineGrid({
           {(() => {
             const unassignedEvents = grouped.get(UNASSIGNED_MEMBER.id) ?? [];
             if (unassignedEvents.length === 0) return null;
+            const { lanes, laneCount } = assignLanes(unassignedEvents);
+            const rowHeight = rowHeightForLanes(laneCount);
             return (
               <DroppableTimelineRow
                 teamMember={UNASSIGNED_MEMBER}
                 startDate={startDate}
                 daysShown={daysShown}
                 isLast
+                rowHeight={rowHeight}
               >
                 {unassignedEvents.map((event) => (
                   <TimelineTaskBlock
@@ -346,6 +422,9 @@ export function TimelineGrid({
                     startDate={startDate}
                     daysShown={daysShown}
                     isSelected={isTaskSelected(event.id)}
+                    laneIndex={lanes.get(event.id) ?? 0}
+                    laneCount={laneCount}
+                    rowHeight={rowHeight}
                     onClick={handleEventClick}
                     onContextMenu={handleContextMenu}
                     onResize={handleResize}
--- a/src/app/(dashboard)/calendar/_components/timeline/timeline-row.tsx
+++ b/src/app/(dashboard)/calendar/_components/timeline/timeline-row.tsx
@@ -17,6 +17,8 @@ interface TimelineRowProps {
   startDate: Date;
   daysShown: number;
   isLast?: boolean;
+  /** Optional row height override — used by lane stacking when overlaps push the row taller. */
+  rowHeight?: number;
   children?: ReactNode;
 }
 
@@ -27,6 +29,7 @@ export function TimelineRow({
   startDate,
   daysShown,
   isLast = false,
+  rowHeight,
   children,
 }: TimelineRowProps) {
   const fullName = `${teamMember.firstName} ${teamMember.lastName}`.trim() || "Unknown";
@@ -38,7 +41,7 @@ export function TimelineRow({
     <div
       className="flex group transition-colors duration-150"
       style={{
-        height: TIMELINE_ROW_HEIGHT,
+        height: rowHeight ?? TIMELINE_ROW_HEIGHT,
         borderBottom: isLast ? "none" : "1px solid rgba(255,255,255,0.05)",
         background: "transparent",
       }}
--- a/src/app/(dashboard)/calendar/_components/timeline/timeline-task-block.tsx
+++ b/src/app/(dashboard)/calendar/_components/timeline/timeline-task-block.tsx
@@ -16,6 +16,12 @@ interface TimelineTaskBlockProps {
   daysShown: number; // number of visible day columns
   isSelected?: boolean; // selected via click or multi-select
   isGhost?: boolean; // ghost preview for cascade/auto-schedule
+  /** Vertical lane index for stacking overlapping events (0-based) */
+  laneIndex?: number;
+  /** Total number of lanes used in this row */
+  laneCount?: number;
+  /** Total row height in px — block divides this evenly across laneCount */
+  rowHeight?: number;
   onClick?: (event: InternalCalendarEvent) => void;
   onContextMenu?: (
     event: InternalCalendarEvent,
@@ -63,6 +69,9 @@ export function TimelineTaskBlock({
   daysShown,
   isSelected = false,
   isGhost = false,
+  laneIndex = 0,
+  laneCount = 1,
+  rowHeight = TIMELINE_ROW_HEIGHT,
   onClick,
   onContextMenu,
   onResize,
@@ -145,7 +154,19 @@ export function TimelineTaskBlock({
     return { leftPercent: newLeft, widthPercent: newWidth };
   }, [resizeState, leftPercent, widthPercent, daysShown]);
 
-  const blockHeight = TIMELINE_ROW_HEIGHT - 16; // 56px (8px padding top + bottom)
+  // Lane-aware vertical layout. The row reserves 8px top + 8px bottom
+  // padding, then divides the remaining vertical space among laneCount
+  // lanes with a 4px gap between lanes.
+  const VERTICAL_PADDING = 8;
+  const LANE_GAP = 4;
+  const innerHeight = rowHeight - VERTICAL_PADDING * 2;
+  const totalLaneGaps = LANE_GAP * Math.max(laneCount - 1, 0);
+  const perLaneHeight = Math.max(
+    14, // minimum readable lane height
+    Math.floor((innerHeight - totalLaneGaps) / Math.max(laneCount, 1))
+  );
+  const blockHeight = perLaneHeight;
+  const blockTop = VERTICAL_PADDING + laneIndex * (perLaneHeight + LANE_GAP);
 
   // ── Determine if block is narrow ──────────────────────────────────────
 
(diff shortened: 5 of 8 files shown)


src/components/ops/keyboard-shortcuts.tsx
```diff
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
...
+      }
+      // Tag-name fallback for cases where instanceof fails (e.g. cross-realm
+      // events dispatched from portals, iframes, or libraries that re-host the
+      // node). Also covers select elements and standard form controls.
+      const tag = node.tagName?.toUpperCase();
+      if (tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT") {
+        return true;
+      }
+      // Walk up looking for a contenteditable ancestor.
+      if (node instanceof HTMLElement && node.closest("[contenteditable='true'], [contenteditable='']")) {
```
jack-opsapp, line 38: REVIEW PASS :: CONCERN

The contenteditable ancestor walk uses the selector `[contenteditable='true'], [contenteditable='']` which intentionally matches the two most common forms but MISSES `contenteditable='plaintext-only'` (Chromium/WebKit, used by some editors and by Slate.js / draft-js plain-text fallbacks) and any value other than 'true'/'' that the spec treats as truthy via inheritance. A trigger scenario: if a Slate-based or plaintext-only mention/composer is later introduced (or already used in note-composer / inbox reply), pressing '2' inside it would still navigate to /projects.

Suggested tightening: rely on the IDL property instead of the attribute string — replace the selector walk with a loop that climbs parents and checks `(el as HTMLElement).isContentEditable`, OR widen the selector to `[contenteditable]:not([contenteditable='false'])`.

src/components/ops/keyboard-shortcuts.tsx
```diff
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
...
-        target.isContentEditable ||
-        target.closest("[contenteditable]")
-      ) {
+      // Don't capture if user is typing in an input — check both the event
+      // target AND the document's active element. The active-element fallback
+      // catches cases where keystrokes are dispatched after a focus/blur race
+      // or where the event target is the body/window rather than the field
+      // itself (observed when a controlled textarea re-renders mid-keystroke).
+      const target = e.target as Element | null;
+      if (isEditableElement(target) || isEditableElement(document.activeElement)) {
```
jack-opsapp, line 51: REVIEW PASS :: CONCERN

`document.activeElement` returns the SHADOW HOST when focus is inside an open shadow root, not the actually-focused element. If any third-party widget mounted in the dashboard layout (e.g. an embedded chat widget, video player controls, future web components) hosts an input inside a shadow DOM, `isEditableElement(document.activeElement)` will see only the host element (likely a custom-tag div) and return false. `e.target` will likewise be retargeted to the shadow host by the browser, so both branches of the guard fail simultaneously and number-key navigation fires while the user is typing.

Suggested fix: walk shadow roots explicitly, e.g.
```ts
function deepActiveElement(): Element | null {
  let el: Element | null = document.activeElement;
  while (el && (el as any).shadowRoot && (el as any).shadowRoot.activeElement) {
    el = (el as any).shadowRoot.activeElement;
  }
  return el;
}
```
and pass the result to `isEditableElement`.

src/components/ops/keyboard-shortcuts.tsx
```diff
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
...
+      const tag = node.tagName?.toUpperCase();
+      if (tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT") {
+        return true;
+      }
+      // Walk up looking for a contenteditable ancestor.
+      if (node instanceof HTMLElement && node.closest("[contenteditable='true'], [contenteditable='']")) {
+        return true;
+      }
+      return false;
+    }
```
jack-opsapp, line 42: REVIEW PASS :: CONCERN

The fix does not handle focus inside same-origin iframes (rich-text editors like CKEditor/TinyMCE/Quill traditionally use a contenteditable BODY inside an iframe; even if not used today, this listener is global and covers the whole dashboard surface). When focus is in an iframe, `document.activeElement` of the parent document returns the `<iframe>` element itself, which is neither an INPUT/TEXTAREA/SELECT nor contenteditable — so the guard does not bail and the keystroke (which bubbles up to window in the parent if the iframe origin matches and the inner handler doesn't stopPropagation) can fire navigation.

Suggested addition: in `isEditableElement`, treat `HTMLIFrameElement` (and `HTMLObjectElement`/`HTMLEmbedElement` for completeness) as editable when they are the activeElement, since the parent window has no reliable way to introspect the inner doc cross-frame.

src/components/ops/keyboard-shortcuts.tsx
```diff
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
...
-      ) {
+      // Don't capture if user is typing in an input — check both the event
+      // target AND the document's active element. The active-element fallback
+      // catches cases where keystrokes are dispatched after a focus/blur race
+      // or where the event target is the body/window rather than the field
+      // itself (observed when a controlled textarea re-renders mid-keystroke).
+      const target = e.target as Element | null;
+      if (isEditableElement(target) || isEditableElement(document.activeElement)) {
         return;
       }
```
jack-opsapp, line 53: REVIEW PASS :: CONCERN

IME composition is not guarded. While composing CJK input, browsers fire `keydown` with `event.key === 'Process'` and `event.isComposing === true`, but the actual character keys (1-9 etc.) can also fire mid-composition with `isComposing: true`. If a user is composing inside a non-editable element (rare) or if the active element check fails for any of the reasons above, the navigation triggers mid-composition and the IME state is lost.

More importantly, even with the editable check working, a user typing romaji that maps to digits (e.g. some kana input modes) could trigger this. Lower severity than the others — flagging as defense-in-depth.

Suggested addition: early-return when `e.isComposing || e.keyCode === 229`.

src/components/ops/keyboard-shortcuts.tsx
```diff
@@ -19,15 +19,36 @@ export function KeyboardShortcuts() {
   const router = useRouter();
 
   useEffect(() => {
```
jack-opsapp, line 21: REVIEW PASS :: NITPICK

Component-level note (anchoring here since line 16 is outside the diff): the JSDoc above this useEffect advertises a `?` shortcut to show keyboard shortcuts help, but no handler for `?` exists in this file. Either the doc is stale or the feature was lost — verify whether help-overlay invocation lives elsewhere and clean up the doc accordingly.

jack-opsapp:
REVIEW PASS (bug 916ca622) :: 5 concerns (0 blockers, 4 concerns, 1 nitpick)

STATUS :: ALLOW MERGE

The fix correctly resolves the reported regression: number-key navigation now bails when either `e.target` or `document.activeElement` is editable, with a tag-name fallback for cross-realm cases and a contenteditable ancestor walk. Verified in `src/components/ops/keyboard-shortcuts.tsx` — the original failure mode (controlled textarea re-render mid-keystroke producing a non-textarea `e.target`) is now caught by the activeElement branch.

No blockers identified. Posted four CONCERNs as defense-in-depth that are out of scope for this PR but should be tracked:

- contenteditable selector misses `plaintext-only` and inherited variants — switch to IDL `isContentEditable` walk
- `document.activeElement` returns shadow host, not focused element inside open shadow roots
- iframe-hosted editors (CKEditor/TinyMCE-style) treated as non-editable since the active element is the `<iframe>` itself
- IME composition not guarded — `e.isComposing` early-return recommended

Plus one NITPICK on stale JSDoc for the unimplemented `?` help shortcut.

None of these reproduce the reported QA scenario (project Notes textarea), so they should not block this hotfix from shipping. Recommend a follow-up PR to harden `isEditableElement` against shadow DOM and iframes, since the listener is global and applies to the entire dashboard surface.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 51: REVIEW PASS :: CONCERN

Bug 94977186 reports two distinct symptoms: (1) 'inputs inaccessible' AND (2) 'submission fails'. The fix in keyboard-shortcuts.tsx only addresses (1) — keystrokes being intercepted by global number-key navigation. There is no change that explains why createProject submission would fail. The fix_notes correctly acknowledges this assumption: 'If this is in fact a different root cause, ask reviewer to repro with a DOM snapshot and re-open.'

Suggested action: do not close 94977186 as 'verified' until QA confirms submission works post-fix. The submission-fail symptom may be a separate bug (e.g. companyId missing from auth store, RLS violation on project insert, or createProject mutation hook regression) that the keystroke fix masks but does not resolve. Recommend keeping bug status as 'in_progress' pending repro.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 42: REVIEW PASS :: CONCERN

If the diagnosed root cause is real (event target resolves to a parent / body when a controlled textarea re-renders mid-keystroke), then EVERY other global keydown listener in the dashboard that uses the same weaker check is still vulnerable to the same race:

- src/components/layouts/top-bar.tsx L115-128 — Cmd+Z undo (only checks tagName/isContentEditable on e.target, NOT document.activeElement).
- src/components/layouts/quick-actions-tab.tsx L24-36 — bare letter 'q' shortcut. Typing the letter q in the project name field could STILL trigger the QA drawer if e.target re-resolves to body during a remount.
- src/components/layouts/notifications-tab.tsx L37-49 — bare letter 'n' shortcut. Same risk for the letter n.
- src/components/ops/command-palette.tsx L106-127 — backslash shortcut, also weaker check on e.target only.

If the QA reporter typed a word containing q or n, the project-name input would lose focus to a side drawer, completing the 'inputs inaccessible' symptom independently of the number-key navigation. The current fix does not patch these.

Suggested fix: extract isEditableElement into a shared util (e.g. src/lib/utils/keyboard.ts) and replace all four call sites so the hardening is uniform.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 38: REVIEW PASS :: NITPICK

The contenteditable closest() walk at L38 is gated on `node instanceof HTMLElement`, so for the cross-realm case the comment at L30-32 calls out (events from portals / iframes), the walk is skipped — only the tag-name fallback would fire, and only if the cross-realm node IS itself an INPUT/TEXTAREA/SELECT. A cross-realm contenteditable div would slip through.

In practice this is unlikely to bite anyone, but if cross-realm protection is the stated reason, consider replacing the second instanceof guard with a duck-typed `typeof (node as HTMLElement).closest === 'function'` check, or just calling `node.closest?.(...)` since closest is on Element across realms.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 51: REVIEW PASS :: NITPICK

Minor consequence: `document.activeElement` falls back to `<body>` when nothing is focused, so isEditableElement(activeElement) returns false there (good). But after the new check, if a user opens the keyboard-shortcut help (?) or any global hotkey while a hidden offscreen input retains focus (e.g. an autofocus search field elsewhere on the dashboard), number-key navigation will be silently swallowed even though the user thinks nothing is focused. Behaviorally this is safer than before, but worth noting in the bug closeout — QA may legitimately report 'number shortcuts no longer work' on a page where an input has stale focus.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 53: REVIEW PASS :: BLOCKER

This commit message and fix_notes attribute the fix to bug 916ca622 (project Notes textarea). Bug 8665d78e is a *separate* report whose reproduction steps are: "Navigate to Clients page → Click New client → Attempt to type first name → Observe input fields inaccessible." The fixer flagged this as "likely same root cause" with no DOM snapshot, no console errors, and no network errors to back that up.

The handler defined here only intercepts: digits 1–8, Cmd+Shift+P, Cmd+Shift+C, and Cmd+B. Typing the letters J-O-H-N into a focused name input never matches any of those branches and never reaches `e.preventDefault()`, regardless of whether the editable-element check passes. Even before this patch, focus on the name field would set `e.target` to the actual `<input>`, so `target instanceof HTMLInputElement` was already true and the handler bailed. So for the reported symptom (first-name field inaccessible) this fix is a no-op.

Suggested fix: do not auto-close 8665d78e on the back of the 916ca622 patch. Either (a) reproduce the New Client Form symptom with a DOM snapshot before merging, or (b) split this bug back to triage with `requires_human_review: true` and reason 'no DOM snapshot; symptom (cannot type letters) is not explained by the keyboard-shortcut handler'. The fix_notes already contemplates this: 'If a separate root cause is later confirmed, ask reviewer to attach a DOM snapshot and re-open.' That should happen before merge, not after.

src/components/layouts/sidebar.tsx
```diff
@@ -274,14 +274,23 @@ export function Sidebar() {
       <aside
         onMouseEnter={() => { if (!isMobileView) setHoverExpanded(true); }}
         onMouseLeave={() => { if (!isMobileView) setHoverExpanded(false); }}
+        aria-hidden={isMobileView && !isMobileOpen ? "true" : undefined}
```
jack-opsapp, line 277: REVIEW PASS :: BLOCKER

`aria-hidden="true"` is set on the `<aside>` when the mobile drawer is closed, but the sidebar still contains real focusable elements (every NavItemButton, the avatar dropdown trigger, etc.). `pointer-events-none` blocks pointer interactions, but it does NOT remove elements from the keyboard tab order. A keyboard user on mobile (or anyone with a Bluetooth keyboard / external switch) can tab into the off-screen drawer, focus controls they cannot see, and trigger navigation. This is also a WCAG 4.1.2 / ARIA 1.1 violation: `aria-hidden` MUST NOT contain focusable descendants — Chrome / axe / Lighthouse will flag this as an error.

Suggested fix: also add the `inert` attribute when closed (it removes focusability AND aria-hides in one shot, and is supported in all browsers OPS targets). Replace the `aria-hidden` line with:

```tsx
{...(isMobileView && !isMobileOpen ? { inert: "" as unknown as undefined } : {})}
```

(or use the boolean prop once the project's React types support it). `inert` covers both pointer + keyboard + AT correctly and lets you drop the `pointer-events-none` / `aria-hidden` machinery.

jack-opsapp:
REVIEW PASS (bug 94977186) :: 3 concerns (0 blockers, 2 concerns, 1 nitpick (1 of 3 marked NITPICK is shown as nitpick))

STATUS :: ALLOW MERGE

The keyboard-shortcuts.tsx hardening is correct and safe — it strictly tightens the global navigation listener with no regression risk to project creation. No blockers.

Concerns to address (do NOT block this PR, but track as follow-ups):

- **Symptom-vs-fix mismatch**: bug 94977186 lists 'submission fails' as a co-symptom. This fix only stops keystroke interception. Verify with a real repro that createProject succeeds before closing the bug; if submission still fails, re-open with a separate root cause (auth/companyId, RLS, mutation hook).
- **Sibling listeners not patched**: top-bar.tsx (Cmd+Z), quick-actions-tab.tsx (Q), notifications-tab.tsx (N), command-palette.tsx (\\) all retain the old weaker target-only check. If the diagnosed mid-keystroke re-render race is real, these are still vulnerable — particularly Q and N which would steal focus while typing in the project-name input. Recommend extracting isEditableElement to src/lib/utils/keyboard.ts and patching all four.
- **Cross-realm contenteditable nit**: closest() walk is gated on `instanceof HTMLElement`, so cross-realm contenteditable divs slip past the protection the comment claims to provide. Low real-world risk; trivial to harden if desired.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 38: REVIEW PASS :: NITPICK

The contenteditable selector `[contenteditable='true'], [contenteditable='']` will not match `contenteditable="plaintext-only"` (used by some rich-text libraries) or `contenteditable="inherit"` (rare but valid). Since this whole block is exactly defending against custom rich-text wrappers, prefer either:

- `node.closest("[contenteditable]")` (matches any contenteditable attribute regardless of value) plus an inline check that the resolved value isn't `'false'`, or
- the runtime API: walk up via `closest('[contenteditable]')` and then check `(el as HTMLElement).isContentEditable` to honor inheritance.

Not a blocker on its own, but it's a near-identical regression to the original `target.closest("[contenteditable]")` line this PR replaced — that line was strictly more permissive.

src/components/layouts/sidebar.tsx
```diff
@@ -274,14 +274,23 @@ export function Sidebar() {
...
-          // Mobile: off-screen by default, slide in when open
+          // Mobile: off-screen by default, slide in when open. We pair the
+          // translate with invisible/pointer-events-none so the drawer is
+          // genuinely removed from interaction even if the transform fails
+          // to apply (some legacy mobile browsers / WebView rendering paths
+          // dropped the sidebar back into layout, which left it overlapping
+          // dashboard widgets at 390px).
           isMobileOpen ? "translate-x-0" : "-translate-x-full",
-          "md:translate-x-0"
+          isMobileOpen ? "visible pointer-events-auto" : "invisible pointer-events-none",
```
jack-opsapp, line 290: REVIEW PASS :: CONCERN

The aside still has `transition-all duration-200 ease-out` (line 281). The newly-added `invisible` class is binary — `visibility` flips to `hidden` the instant `isMobileOpen` becomes false, BEFORE the 200ms slide-out (`-translate-x-full`) animation has time to play. Net effect on close: drawer disappears instantly instead of sliding out smoothly. Open-direction is unaffected (visibility: visible applies immediately, then translate animates in).

This is a UX regression. Two options:
1. Add `transition-property: visibility` with a delay equal to the slide duration on close so visibility flips after the slide finishes (`md:transition-[transform,visibility] motion-safe:[transition-delay:0ms,200ms]` etc — fiddly).
2. Better: switch to `inert` (per the BLOCKER above). `inert` does not interfere with CSS transitions because it is not a CSS property — the slide-out animates fully while inertness applies immediately, which is what you want for accessibility anyway.

src/components/ops/keyboard-shortcuts.tsx
jack-opsapp, line 51: REVIEW PASS :: CONCERN

The phone field on the New Client Form (`src/app/(dashboard)/clients/new/page.tsx` lines 231–239) renders a controlled `<Input>` whose `onChange` calls `setValue('phone', formatted)` and runs a re-render every keystroke. This is a near-identical pattern to the controlled-textarea bug 916ca622 cited as the root cause. If the active-element race ever does fire here, digits 1–8 typed into the phone field would now invisibly route the user to /dashboard, /projects, /calendar, /clients, /team, /map, /pipeline, /invoices — losing the entire form's unsaved data with no warning.

The fix as written DOES catch this case via `isEditableElement(document.activeElement)` after the patch, so this CONCERN is preventative not a defect: please add a regression test (jsdom `fireEvent.keyDown` on `window` while `document.activeElement` is the phone input, plus a body-targeted variant) so this doesn't silently regress when someone refactors the handler again. Without a test the same race will return on the next refactor and the user will lose form data again.

src/components/layouts/sidebar.tsx
```diff
@@ -274,14 +274,23 @@ export function Sidebar() {
       <aside
         onMouseEnter={() => { if (!isMobileView) setHoverExpanded(true); }}
         onMouseLeave={() => { if (!isMobileView) setHoverExpanded(false); }}
+        aria-hidden={isMobileView && !isMobileOpen ? "true" : undefined}
         className={cn(
           "fixed left-0 top-0 h-screen z-[45]",
           "border-r border-[rgba(255,255,255,0.06)]",
           "flex flex-col transition-all duration-200 ease-out",
           effectiveCollapsed ? "w-[72px]" : "w-[256px]",
```
jack-opsapp, line 282: REVIEW PASS :: CONCERN

This fix targets a narrative cause ("transforms fail to apply on legacy WebViews") that is unsubstantiated for the reported viewport — the bug reporter said iPhone 14 / 390px, which is iOS Safari, where `-translate-x-full` is solid and has been for years. No console errors / network errors were captured to confirm a transform-paint race.

The likely actual root cause is a state mismatch in `effectiveCollapsed`. Look at line 215:
```
const effectiveCollapsed = isMobileView ? false : !isHoverExpanded;
```
On mobile, `effectiveCollapsed` is forced to `false` so the drawer is always 256px wide — even when `isMobileOpen` is false. Combined with any state where `-translate-x-full` does not fully clear 256px (e.g. a parent with `overflow-x: visible` and an unintended `transform: none` overriding inheritance, or an SSR/hydration paint frame where the off-screen transform has not yet computed), you get a 256px-wide drawer overlapping content.

A more durable fix would also collapse the drawer to `w-[72px]` when closed on mobile, or hard-pin `left: -256px` when closed instead of relying on transforms. The PR does mask the symptom via `visibility: hidden`, but the underlying mismatch in width-while-closed is still there and may resurface in any future scenario where `visibility` gets unset (e.g. a parent injecting `visibility: visible` for tooltips/popovers — Radix and dnd-kit both do this in places).

src/components/layouts/sidebar.tsx
```diff
@@ -274,14 +274,23 @@ export function Sidebar() {
...
+          // genuinely removed from interaction even if the transform fails
+          // to apply (some legacy mobile browsers / WebView rendering paths
+          // dropped the sidebar back into layout, which left it overlapping
+          // dashboard widgets at 390px).
           isMobileOpen ? "translate-x-0" : "-translate-x-full",
-          "md:translate-x-0"
+          isMobileOpen ? "visible pointer-events-auto" : "invisible pointer-events-none",
+          // Restore visibility/interaction at md+ regardless of mobile-open
+          // state so the desktop hover-rail keeps working.
+          "md:translate-x-0 md:visible md:pointer-events-auto"
```
jack-opsapp, line 293: REVIEW PASS :: NITPICK

`aria-hidden={isMobileView && !isMobileOpen ? "true" : undefined}` only resolves correctly after the `useEffect` that sets `isMobileView` runs (post-mount). On a 390px viewport, the first server-rendered + first-paint client render has `isMobileView=false`, so `aria-hidden` evaluates to `undefined` — meaning during the initial paint frame the screen reader sees the off-screen drawer as a live, focusable region. This window is short (one tick) but real, and it also creates a hydration class-string mismatch warning in dev because the `md:visible md:pointer-events-auto` chain depends on `isMobileView` flipping post-mount. Consider initializing `isMobileView` from a `useSyncExternalStore` + `matchMedia` subscription so the first client render already knows the viewport bucket.

jack-opsapp:
REVIEW PASS (bug 8665d78e) :: 3 concerns (1 blocker, 1 concern, 1 nitpick)

STATUS :: BLOCK MERGE

Summary of blockers:
- The fix attributes 8665d78e to the same root cause as 916ca622 (project Notes textarea) without evidence. Bug 8665d78e's reproduction is 'cannot type first name on New Client Form'. Letters typed into a focused name input never match any branch in keyboard-shortcuts.tsx and were already bailed-on by the original `target instanceof HTMLInputElement` check, so the patch is a no-op for the reported symptom. fix_notes itself says 'If a separate root cause is later confirmed, ask reviewer to attach a DOM snapshot and re-open' — that needs to happen before merge, not after.

Other notes:
- CONCERN: phone field uses a controlled Input with re-render-per-keystroke that matches the 916ca622 race condition. The patch does cover this case (active-element fallback), but there is no regression test guarding it; add one before this gets refactored away again.
- NITPICK: `closest("[contenteditable='true'], [contenteditable='']")` is stricter than the original `closest('[contenteditable]')` it replaces — misses `contenteditable='plaintext-only'` and `'inherit'`. Use `closest('[contenteditable]')` plus an isContentEditable check on the resolved element.

jack-opsapp:
REVIEW PASS (bug 5b32195c) :: 3 concerns (1 blocker, 2 concerns, 1 nitpick)

STATUS :: BLOCK MERGE

Summary of blockers:
- `aria-hidden="true"` is applied to the `<aside>` while it still contains focusable nav items / dropdown triggers. `pointer-events-none` does NOT remove keyboard focusability, so a Bluetooth-keyboard / external-switch user on mobile can tab into the off-screen drawer and trigger navigation. This is also a WCAG 4.1.2 / ARIA 1.1 violation that axe / Lighthouse will flag. Use the `inert` attribute instead — it covers focus + pointer + AT in one go and lets the existing `transition-all duration-200` slide-out animate properly (visibility:hidden currently snaps the drawer away before the close-slide runs).

Other notes:
- The fix's stated root cause (transforms failing on legacy WebViews at iPhone 14 / 390px) is unsubstantiated. The actual cause is more likely the `effectiveCollapsed = isMobileView ? false : !isHoverExpanded` rule keeping the drawer at 256px wide while closed — a more durable fix would also collapse width or hard-pin `left:-256px` instead of relying on transforms.
- SSR/hydration: `isMobileView` initializes to false, so the first paint on a 390px viewport leaks an unhidden drawer for one tick and may emit a hydration class-string warning. Initialize from a `matchMedia` `useSyncExternalStore` subscription.

PR #23, opened 2026-04-28
Nightly Bug Sweep — web — 2026-04-28 — Batch 01

jack-opsapp:
## Summary

Nightly bug sweep covering 6 web bugs (5 fixed, 1 escalated for human review).

`npm run type-check` passes on every commit. `npm run build` fails on this branch AND on `main` with the same pre-existing error: `supabaseUrl is required` during page-data collection at `/api/admin/whats-new/categories` — environmental, unrelated to these fixes.

---

### fix(bug-fd1f8302) — Inbox error screen redesigned to v2 spec  *(ui_issue)*
Old error.tsx used legacy Mohave display tokens and centered layout, conflicting with spec v2 (Cake Mono Light display, left-only alignment, glass surfaces, tactical voice). Rebuilt with `font-cakemono font-light uppercase` 28px title, left-aligned column, `glass-surface` diagnostic panel, status-rose pulse dot, `// SYS :: ROUTE FAULT` voice, `[MSG] / [DIGEST] / [TIME]` bracketed metadata.

### fix(bug-08571812) — portal_branding visibility-column save failure  *(bug · urgent)*
`SAVE BRANDING` was failing entirely with "Could not find the 'show_descriptions' column of 'portal_branding' in the schema cache" because migration 044 (visibility columns: show_quantities/unit_prices/line_totals/descriptions/tax/discount) hasn't been applied to the target Postgres / PostgREST schema cache. The error blocked every branding edit including logo / accent / template / theme. Fix: detect the schema-cache missing-column error in `updateBranding`, strip every visibility-override column from the payload, and retry. Core branding now saves; a console warning surfaces that migration 044 is missing.

### fix(bug-ac030a6c) — Subscription lockout bypassed (no payment info)  **COMPLEX**  *(bug · subscription·payment·lockout)*
Pete Mitchell could click `UPGRADE TO STARTER` and be granted the upgraded UX without entering payment info. Root cause: `/api/stripe/subscribe` accepted requests with no `paymentMethodId`, created an `incomplete` Stripe sub (mapper returns null → `subscription_status` untouched), but still wrote `subscription_plan` / `subscription_period` / `max_seats` / `subscription_end` / `subscription_ids_json` and surfaced a "Subscribed!" toast. Server now requires either a passed `paymentMethodId` OR an existing `default_payment_method` on the Stripe customer; rejects with HTTP 402 `code:"payment_method_required"` otherwise. Frontend `UpgradeModal` intercepts the 402 and routes the user to Settings → Billing instead of showing success.

### fix(bug-50248c4d) — Stripe price id "no such price"  **COMPLEX**  *(bug · subscription)*
`UPGRADE TO STARTER` surfaced raw Stripe error "no such price id". Root cause: `STRIPE_PRICE_STARTER_*` env var points at a price id that doesn't exist in the connected Stripe account; `stripe.subscriptions.create` throws `StripeInvalidRequestError code=resource_missing param=items[0][price]`. Fix: catch the error type explicitly, return a user-friendly "This plan is currently unavailable. Contact support…" with `code:"price_not_found"` + the original Stripe message attached for diagnostics. Other Stripe invalid-request errors return 400 with `code` + `param`. The misconfigured env var still needs operator action; this fix surfaces the misconfiguration rather than crashing.

### fix(bug-bffbad6b) — Calibration interview crash: getSnapshot infinite loop  *(bug · urgent · crash-shaped)*
Starting an interview crashed with "The result of getSnapshot should be cached to avoid an infinite loop" → Maximum update depth exceeded. Root cause: `ProgressBar` (`ai-intake-interview.tsx:71`) subscribes to `selectCategoryProgress` and `selectProgress`, both of which build a fresh object on every call. Zustand v5 wraps subscriptions in `useSyncExternalStore`, which compares snapshot reference equality — a new object every render means React re-fires unbounded. Fix: wrapped both object-returning selectors in `useShallow` from `zustand/react/shallow` at the call site. Primitive selectors (`selectCurrentCategory`, etc.) untouched.

---

### bug-1c468d20 — Uniform layouts/toggles across data tabs  *(ui_issue · high)*
Reporter asks for unified list/grid view toggles across Clients / Projects / Inventory / Invoices / Estimates. Each page currently has bespoke layouts (e.g. Projects has ProjectCanvas + ProjectSpreadsheet + StageStack). This is a multi-PR design+engineering initiative requiring a shared DataView/ViewToggle primitive, per-entity column maps, preferences persistence, and migration of every existing layout — not a nightly bugfix. Escalated for product/design alignment.

---

## Test plan

- [ ] Visit `/inbox` and trigger a route error — verify new tactical error screen renders (left-aligned, Cake Mono title, glass panel, "// SYS :: ROUTE FAULT")
- [ ] Settings → Portal Branding → change logo/template and `SAVE BRANDING` — verify save succeeds even when migration 044 is unapplied (visibility overrides silently skip; console warns)
- [ ] Settings → Subscription → click `UPGRADE TO STARTER` as a user with no card on file — verify 402 → "Payment method required" toast with "Open Billing" action
- [ ] With the misconfigured Stripe price env, click `UPGRADE TO STARTER` — verify "This plan is currently unavailable…" message instead of "no such price id"
- [ ] Calibration → start interview — verify no infinite-loop / Maximum-update-depth crash; ProgressBar renders with category segments

### DIFF
--- a/src/app/api/stripe/subscribe/route.ts
+++ b/src/app/api/stripe/subscribe/route.ts
@@ -87,7 +87,15 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
     const priceId = getPriceId(plan, period);
     if (!priceId) {
       console.error(`[stripe/subscribe] Missing price env var for ${plan}_${period}`);
-      return NextResponse.json({ error: "Price configuration not found" }, { status: 500 });
+      return NextResponse.json(
+        {
+          error: `Price configuration not found for ${plan} ${period}. Contact support.`,
+          code: "price_env_missing",
+          plan,
+          period,
+        },
+        { status: 500 }
+      );
     }
 
     const supabase = getServiceRoleClient();
@@ -300,6 +308,36 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
       clientSecret,
     });
   } catch (error) {
+    // Stripe errors carry a `code` field (e.g. "resource_missing") and a
+    // `param` (e.g. "items[0][price]") that pinpoint config issues. Surface
+    // them cleanly so users see a useful message instead of "no such price"
+    // and so logs identify the misconfigured env var.
+    if (error instanceof Stripe.errors.StripeInvalidRequestError) {
+      const code = error.code ?? "stripe_invalid_request";
+      const param = error.param ?? null;
+      console.error(
+        `[stripe/subscribe] StripeInvalidRequestError code=${code} param=${param ?? "?"} msg=${error.message}`
+      );
+      if (code === "resource_missing" && (param === "price" || param?.includes("price"))) {
+        return NextResponse.json(
+          {
+            error:
+              "This plan is currently unavailable. Contact support so we can update the pricing configuration.",
+            code: "price_not_found",
+            stripeMessage: error.message,
+          },
+          { status: 500 }
+        );
+      }
+      return NextResponse.json(
+        {
+          error: error.message,
+          code,
+          param,
+        },
+        { status: 400 }
+      );
+    }
     const message = error instanceof Error ? error.message : "Failed to create subscription";
     console.error("[stripe/subscribe] Error:", error);
     return NextResponse.json({ error: message }, { status: 500 });
--- a/src/app/api/stripe/subscribe/route.ts
+++ b/src/app/api/stripe/subscribe/route.ts
@@ -187,6 +187,37 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
       await stripe.customers.update(stripeCustomerId, {
         invoice_settings: { default_payment_method: paymentMethodId },
       });
+    } else {
+      // Lockout enforcement — refuse to create a subscription without a
+      // confirmed payment method. Without this guard, a UI bug or stale
+      // upgrade modal can call /subscribe with no paymentMethodId, Stripe
+      // creates an `incomplete` subscription that never becomes active, but
+      // the company row gets `subscription_plan` written + a success toast
+      // surfaces — granting the user the upgraded tier UX before any money
+      // changes hands.
+      //
+      // If a paymentMethodId was not passed, require that the Stripe customer
+      // already has a default payment method on file (set via SetupIntent in
+      // /settings/billing or attached on a prior subscribe call). Otherwise
+      // reject with 402 so the frontend can route the user to add a card.
+      const customer = await stripe.customers.retrieve(stripeCustomerId);
+      if (customer.deleted) {
+        return NextResponse.json(
+          { error: "Stripe customer was deleted" },
+          { status: 500 }
+        );
+      }
+      const defaultPm = customer.invoice_settings?.default_payment_method;
+      if (!defaultPm) {
+        return NextResponse.json(
+          {
+            error:
+              "A payment method is required to subscribe. Add a card in Settings → Billing first.",
+            code: "payment_method_required",
+          },
+          { status: 402 }
+        );
+      }
     }
 
     // Create subscription.
--- a/src/components/settings/subscription-tab.tsx
+++ b/src/components/settings/subscription-tab.tsx
@@ -105,7 +105,26 @@ function UpgradeModal({
         }),
       });
       const data = await res.json();
-      if (!res.ok) throw new Error(data.error ?? "Subscription failed");
+      if (!res.ok) {
+        // 402 from /api/stripe/subscribe means the company has no payment
+        // method on file. Surface a specific, actionable toast that routes
+        // the user to Settings → Billing instead of the generic failure
+        // message — and bail BEFORE showing the success toast.
+        if (res.status === 402 || data?.code === "payment_method_required") {
+          toast.error("Payment method required", {
+            description:
+              "Add a card in Settings → Billing, then try the upgrade again.",
+            action: {
+              label: "Open Billing",
+              onClick: () => {
+                window.location.assign("/settings?tab=billing");
+              },
+            },
+          });
+          return;
+        }
+        throw new Error(data.error ?? "Subscription failed");
+      }
       toast.success(t("subscription.toast.subscribed"));
       onClose();
     } catch (err) {
--- a/src/app/(dashboard)/error.tsx
+++ b/src/app/(dashboard)/error.tsx
@@ -2,7 +2,7 @@
 
 import { useEffect, useState } from "react";
 import { useRouter } from "next/navigation";
-import { AlertCircle, RefreshCw, LayoutDashboard, ChevronDown } from "lucide-react";
+import { RefreshCw, LayoutDashboard, ChevronDown } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { OpsLockup } from "@/components/brand";
 import { cn } from "@/lib/utils/cn";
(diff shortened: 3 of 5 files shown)


src/app/api/stripe/subscribe/route.ts
```diff
@@ -269,6 +308,36 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+      );
+      if (code === "resource_missing" && (param === "price" || param?.includes("price"))) {
+        return NextResponse.json(
+          {
+            error:
+              "This plan is currently unavailable. Contact support so we can update the pricing configuration.",
+            code: "price_not_found",
+            stripeMessage: error.message,
+          },
+          { status: 500 }
```
jack-opsapp, line 329: REVIEW PASS :: BLOCKER

The fix surfaces a friendlier error message but does NOT actually fix the bug as reported. The reporter clicked UPGRADE TO STARTER and got 'no such price id' — after this PR they will still be unable to subscribe; they just see 'This plan is currently unavailable. Contact support so we can update the pricing configuration.' instead. The fix_notes acknowledge this: 'the actual env var still needs to be corrected (or the corresponding Stripe price restored) — that is operator action, not a code change.'

Concrete scenario: Pete Mitchell (the bug reporter, admin at MAVERICK PROJECTS LTD) revisits Settings → Subscription tomorrow, clicks UPGRADE TO STARTER, and is still unable to subscribe. The bug stays open from his perspective.

Suggested fix: this is a code-only band-aid. Either (a) keep this PR, mark the bug 'requires operator action' with a tracked next step on env var correction (STRIPE_PRICE_STARTER_MONTHLY/ANNUAL in the deployed env), or (b) source price ids from a database table (e.g. subscription_plans) so they can be corrected without a redeploy. The current state is 'fixed the error UX, did not fix the user-blocking bug' — this needs to be flagged so the operator action does not get lost.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -269,6 +308,36 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+      console.error(
+        `[stripe/subscribe] StripeInvalidRequestError code=${code} param=${param ?? "?"} msg=${error.message}`
+      );
+      if (code === "resource_missing" && (param === "price" || param?.includes("price"))) {
+        return NextResponse.json(
+          {
+            error:
+              "This plan is currently unavailable. Contact support so we can update the pricing configuration.",
+            code: "price_not_found",
+            stripeMessage: error.message,
```
jack-opsapp, line 327: REVIEW PASS :: CONCERN

`stripeMessage: error.message` exposes Stripe's raw error text — which for resource_missing reliably contains the literal price ID, e.g. 'No such price: "price_1PqXXXXXXXXXXX"'. That price ID gets shipped to the browser response body and visible to anyone with DevTools. While Stripe price IDs are not strictly secret, broadcasting the misconfigured ID to every user who clicks Upgrade is needless data leakage and gives an attacker a useful enumeration vector against the connected Stripe account (e.g. confirms the account is in test vs. live mode, reveals naming patterns).

Concrete scenario: any unauthenticated competitor who can sign up for a trial clicks Upgrade, opens the Network tab, and reads the production Stripe price ID + 'No such price' message — useful intel for social engineering or impersonation.

Suggested fix: do not return `stripeMessage` to the client. Log it server-side (already happens on line 319) and pass an internal correlation id back instead, or just omit it entirely — the user-facing copy already tells them to contact support.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -269,6 +308,36 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+    // `param` (e.g. "items[0][price]") that pinpoint config issues. Surface
+    // them cleanly so users see a useful message instead of "no such price"
+    // and so logs identify the misconfigured env var.
+    if (error instanceof Stripe.errors.StripeInvalidRequestError) {
+      const code = error.code ?? "stripe_invalid_request";
+      const param = error.param ?? null;
+      console.error(
+        `[stripe/subscribe] StripeInvalidRequestError code=${code} param=${param ?? "?"} msg=${error.message}`
+      );
+      if (code === "resource_missing" && (param === "price" || param?.includes("price"))) {
```
jack-opsapp, line 321: REVIEW PASS :: CONCERN

`param?.includes("price")` is overly broad. It will match any param string containing the substring 'price' — `pricing_table`, `metadata[price_tag]`, `tax_rate.price_includes_tax`, etc. — and silently route those into the 'plan unavailable' branch even when the underlying error has nothing to do with the upgrade price ID. Combined with code === 'resource_missing', this is the case where Stripe says 'no such pricing_table id' and the user sees 'This plan is currently unavailable' which is misleading.

Concrete scenario: future code adds `pricing_table` lookup or `tax_rates` lookup to this route. A typo in that env var triggers resource_missing with a 'pricing_*' or 'tax_rates[0].*' param, and the user gets the 'plan unavailable, contact support' message — dispatching support down the wrong path.

Suggested fix: use `param === "price" || param.startsWith("items[") && param.endsWith("[price]")` (matching only the legitimate Stripe v20+ shape `items[N][price]`), or just use `param === "price" || /^items\[\d+\]\[price\]$/.test(param)`.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -269,6 +308,36 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+              "This plan is currently unavailable. Contact support so we can update the pricing configuration.",
+            code: "price_not_found",
+            stripeMessage: error.message,
+          },
+          { status: 500 }
+        );
+      }
+      return NextResponse.json(
+        {
+          error: error.message,
```
jack-opsapp, line 334: REVIEW PASS :: CONCERN

The 400 fallback returns `error: error.message` directly to the client. `StripeInvalidRequestError.message` can include sensitive details — internal IDs, account-mode hints (test vs live), parameter values that originate from server config. This widens the data-exposure surface; pre-fix it was the same (the catch-all returned `error.message` too) but this PR is the natural place to tighten it.

Concrete scenario: a future code path passes a misconfigured customer id, gets resource_missing with param='customer', falls into this 400 branch — the response body contains the Stripe customer id (`cus_XXXXX`) tied to the company. That id is then exposed in the user's network tab.

Suggested fix: return only a normalized error code + a generic user-safe string. Keep the detailed message in server logs only.

src/app/api/stripe/subscribe/route.ts
jack-opsapp, line 329: REVIEW PASS :: NITPICK

Status 500 for `price_not_found` semantically misclassifies a configuration issue as a server failure. The frontend (`subscription-tab.tsx:108-126`) only special-cases 402 (`payment_method_required`); everything else falls into the generic 'subscribeFailed' toast with `description: data.error`. So the friendly long error string ('This plan is currently unavailable. Contact support so we can update the pricing configuration.') ends up as a toast description, and the toast title remains the generic 'Subscription failed' from the dictionary — partial UX win.

Concrete scenario: user clicks UPGRADE, sees toast with title 'Subscription failed' and a long description in the body. Slightly buried.

Suggested fix: special-case `code === 'price_not_found'` in subscription-tab.tsx `handleSubscribe` to render a toast with a clearer title (e.g. 'PLAN UNAVAILABLE') and just the action 'Contact support'. Optionally use 503 (Service Unavailable) instead of 500 to convey 'temporary, contact us' semantics.

jack-opsapp:
REVIEW PASS (bug 50248c4d — pricing/price id) :: 4 concerns (1 blocker, 2 concerns, 1 nitpick)

STATUS :: BLOCK MERGE

Summary of blockers:
- The fix does not actually unblock the reporter. It surfaces a friendlier error message but the user still cannot subscribe — `STRIPE_PRICE_STARTER_*` env var still points at a missing price id. The fix_notes acknowledge this. Either ship this PR with the operator action explicitly tracked (correct the env var / restore the Stripe price), or move price-id source-of-truth to a database table so it can be repaired without redeploy. As-is, the bug remains user-blocking after merge.

Other concerns:
- `stripeMessage` leaks raw Stripe error text (incl. the literal price id) to the client response body
- `param?.includes("price")` matches unrelated params (`pricing_table`, `metadata[price_tag]`, etc.) — tighten to a regex or strict equality on `items[N][price]`
- 500 status + missing frontend special-case for `price_not_found` means the friendly message lands as a toast description under the generic 'Subscription failed' title

src/app/api/stripe/subscribe/route.ts
```diff
@@ -187,6 +187,37 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+      }
+      const defaultPm = customer.invoice_settings?.default_payment_method;
+      if (!defaultPm) {
+        return NextResponse.json(
+          {
+            error:
+              "A payment method is required to subscribe. Add a card in Settings → Billing first.",
+            code: "payment_method_required",
+          },
+          { status: 402 }
```
jack-opsapp, line 226: REVIEW PASS :: CONCERN

The lockout has a hole when the customer has a default_payment_method on file but it's expired/declined/inactive. Stripe will still let `subscriptions.create` succeed (it returns an `incomplete` sub when default_incomplete + the card later fails), `mapStripeStatus("incomplete") === null` leaves status untouched, but lines 264-292 still write subscription_plan / max_seats / subscription_period / subscription_end / subscription_ids_json AND the route returns 200 → frontend shows the success toast and dismisses the modal. Same UX-grant-without-payment shape as the original bug, just with a narrower trigger (user must have an old card on file).

Concrete scenario: company adds a card during a previous starter attempt that later expired in Stripe. Card stays as customer.invoice_settings.default_payment_method (Stripe doesn't auto-clear). User clicks Upgrade to Team. The lockout passes. Stripe creates incomplete sub. Company row gets subscription_plan=team, max_seats=5 written. Trial user with 8 seated employees is now over-quota with no money taken.

Suggested fix: don't trust default_payment_method blindly — either (a) skip this fast-path entirely and always require a fresh paymentMethodId on every /subscribe call (the billing tab already has SetupIntent for adding cards), or (b) gate the subscription_plan/max_seats/period/end writes on lines 264-273 the same way subscription_status is gated (only write when mappedStatus !== null), so an `incomplete` Stripe sub doesn't grant tier UX.

src/components/settings/subscription-tab.tsx
```diff
@@ -105,7 +105,26 @@ function UpgradeModal({
...
-      if (!res.ok) throw new Error(data.error ?? "Subscription failed");
+      if (!res.ok) {
+        // 402 from /api/stripe/subscribe means the company has no payment
+        // method on file. Surface a specific, actionable toast that routes
+        // the user to Settings → Billing instead of the generic failure
+        // message — and bail BEFORE showing the success toast.
+        if (res.status === 402 || data?.code === "payment_method_required") {
+          toast.error("Payment method required", {
+            description:
+              "Add a card in Settings → Billing, then try the upgrade again.",
```
jack-opsapp, line 116: REVIEW PASS :: NITPICK

Hardcoded English copy ("Payment method required", "Add a card in Settings → Billing, then try the upgrade again.", "Open Billing") in a file that otherwise uses `useDictionary("settings")` for every visible string (e.g. line 128's `t("subscription.toast.subscribed")`, line 131's `t("subscription.toast.subscribeFailed")`). A Spanish-locale user hitting the lockout will see English copy in the toast. Add new keys to `src/i18n/dictionaries/{en,es}/settings.json` (e.g. `subscription.toast.paymentMethodRequired`, `subscription.toast.paymentMethodRequiredDescription`, `subscription.toast.openBilling`) and reference via `t(...)`.

src/components/settings/subscription-tab.tsx
```diff
@@ -105,7 +105,26 @@ function UpgradeModal({
...
+            action: {
+              label: "Open Billing",
+              onClick: () => {
+                window.location.assign("/settings?tab=billing");
```
jack-opsapp, line 120: REVIEW PASS :: NITPICK

`window.location.assign("/settings?tab=billing")` is a hard navigation — it triggers a full page reload, blows away the TanStack Query cache, the Zustand stores (auth, permissions, modals), and the user has to re-hydrate everything before they can add a card. The user is already on `/settings`, so the change here is purely a tab swap. Use `useRouter().push("/settings?tab=billing")` (or `useSearchParams` + setter, depending on how the parent reads the tab) so the client-side router handles the tab change without losing state or remounting the entire settings page.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -187,6 +187,37 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+        );
+      }
```
jack-opsapp, line 228: REVIEW PASS :: NITPICK

No `console.log`/audit trail when the 402 fires. Other guard branches in this route log liberally (e.g. line 168 'CAS update failed', line 185 'Lost customer-create race', line 295 'Failed to update subscription fields', line 319 'StripeInvalidRequestError'). A user repeatedly hitting the lockout will be invisible to operators, which makes it harder to spot a UI regression that strips the paymentMethodId from the request body. Add `console.warn(\`[stripe/subscribe] Blocked subscribe — no payment method on file for company \${companyId}\`)` before this 402 return so the case is observable in Vercel logs.

jack-opsapp:
REVIEW PASS (bug ac030a6c — subscription lockout) :: 4 concerns (0 blockers, 1 concern, 3 nitpicks)

STATUS :: ALLOW MERGE

The fix correctly closes the reported bug — clicking Upgrade with zero payment methods on the Stripe customer now returns 402 + an actionable toast instead of writing tier fields and surfacing a success toast.

The one CONCERN (not a blocker, but worth a follow-up) is a residual hole: if the customer has any default_payment_method on file (e.g. an expired card from a prior subscribe attempt — Stripe doesn't auto-clear these), the lockout passes, Stripe creates an `incomplete` sub, the mapper returns null for status (correct), but lines 264-273 still write subscription_plan / max_seats / subscription_period / subscription_end / subscription_ids_json and the route returns 200. The success toast fires and the user sees the upgraded tier UX without payment confirming. The cleanest fix is to gate those writes on `mappedStatus !== null` the same way subscription_status is gated. Worth tracking as a follow-up bug rather than blocking this merge — the explicit no-card-on-file path that Pete hit is solidly closed.

Nitpicks: hardcoded English copy in the toast (rest of the file uses `useDictionary`), `window.location.assign` causes a full reload when `router.push` would do, and the 402 path has no console.log for observability.

PR #30, opened 2026-04-30
Nightly Bug Sweep — web — 2026-04-30 — Batch 01

jack-opsapp:
## Summary

Nightly batch fix sweep — 10/10 bug reports landed on this branch. All fixes type-check and build clean. Each commit is a single-bug fix with the bug short-id in the subject; full root-cause + edge-case notes are on each `bug_reports` row in the triage DB.

## Bugs fixed

- [ ] **bug-7424ea4f** `bug_reports` — Calendar event badges paint over a busy canvas with frosted-glass-tinted fill; added `frostedBadgeStyle()` helper, applied across Month/Day/Week/Crew bars and the hover popover, and updated `.interface-design/system.md` with the canonical recipe.
- [ ] **bug-cc515384** `bug_reports` — Unscheduled tray is now a dnd-kit droppable; dropping a calendar event on the tray clears `startDate`/`endDate`/`startTime`/`endTime` (recurrence-aware).
- [ ] **bug-a561f726** `bug_reports` — Today cell now uses the same frosted-glass tinted accent (22% alpha over dense glass + backdrop blur) in Month, Week, Crew row, and Crew header — no more invisible 0.06 alpha.
- [ ] **bug-ab3ace6e** `bug_reports` — Universal search palette projects/clients/tasks rows had `value` strings prefixed with the entity UUID, which dropped the cmdk match score below the visibility threshold; switched to user-meaningful searchable text only.
- [ ] **bug-c971a658** `bug_reports` — `/projects` `Add task` was a `toast.info("coming soon")` stub; wired it to `useWindowStore.openWindow({ type: 'create-task', metadata: { projectId } })` and threaded `initialProjectId` through `<CreateTaskForm>`.
- [ ] **bug-1f76cbac** `bug_reports` — Logo loader iframe rendered blank while React+Babel CDN bootstrapped; layered the static `OpsLockup` behind the iframe and fade it out on `iframe.onLoad`. Removed `loading="lazy"` to guarantee an immediate fetch.
- [ ] **bug-1b2942d5** `bug_reports` `[COMPLEX]` — CrewGrid mounted its own nested `<DndContext>` so the unscheduled-tray drag never reached crew rows; introduced `CalendarDragBridgeContext` so descendants register aux handlers, removed the inner DndContext, and switched the unscheduled-task → crew-row branch in `useCrewDnd` from `createMutation` (which produced a duplicate task) to `updateMutation` that sets startDate/endDate/teamMemberIds.
- [ ] **bug-5c267a5e** `bug_reports` — `useUpdateTask`'s `onMutate` updated only the single-task detail cache; calendar/list caches stayed stale until refetch, which is what produced the "badge jumps back, then jumps forward" glitch. Extended `onMutate` to patch every cached `tasks.list`, `tasks.projectTasks`, and `calendar.scheduled` query in place, with full rollback on error.
- [ ] **bug-8620c037** `bug_reports` — Unscheduled tray now docks on the right in every view (was left in Day, right in Week/Month/Crew).
- [ ] **bug-e06445fe** `bug_reports` — Unscheduled tray's Group: Client option resolves the actual client name via `useClients()` and a `clientId -> name` lookup, instead of falling back to `project.title`.

## Test plan

- [ ] Open `/calendar` in any view — calendar event chips/bars/badges render with the new frosted-glass tinted fill; today cell reads as a clear primary-accent highlight without bleeding.
- [ ] Drag a Month event onto the unscheduled rail (left or right collapse state) — event disappears from the calendar and shows up in the tray.
- [ ] Drag an unscheduled task into Crew view onto Jake's row, drop on Wednesday — task gets assigned to Jake and scheduled for Wednesday; the source row in the tray is removed (no duplicate created).
- [ ] Drag any event to a new day — badge stays at the drop target through the server round-trip; no flash to original cell.
- [ ] Switch between Day, Week, Month, Crew — unscheduled tray stays docked on the right in every view.
- [ ] Open `/calendar`, switch the tray's Group dropdown to "Client" — group headings show client names (not project titles).
- [ ] Topbar search → type a project / client / task name — the entity row appears under the matching group.
- [ ] `/projects` spreadsheet → click `Add task` on any row — the create-task floating window opens with that project preselected.
- [ ] `/projects/{id}` (or any first-paint dashboard route while data loads) — logo loader area is never blank; static `OpsLockup` shows immediately, animated chevron fades in once the iframe is ready.

### DIFF
--- a/src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
+++ b/src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
@@ -186,9 +186,7 @@ export function CalendarDndShell({ children }: CalendarDndShellProps) {
       const overData = over.data?.current as
         | { type?: string; day?: Date }
         | undefined;
-      if (!activeData?.type || !overData?.day) return;
-
-      const targetDay = overData.day;
+      if (!activeData?.type) return;
 
       // Recurrence-aware dispatcher
       const dispatchUpdate = async (
@@ -221,6 +219,42 @@ export function CalendarDndShell({ children }: CalendarDndShellProps) {
           }
         );
       };
+
+      // ── Drop on the unscheduled dock → unschedule the event ─────────────
+      // Only existing calendar events can be unscheduled; tasks already in
+      // the tray (`unscheduled-task`) have no schedule to clear.
+      if (overData?.type === "unscheduled-dock") {
+        const calEvent = activeData.event;
+        if (
+          !calEvent ||
+          (activeData.type !== "month-event" &&
+            activeData.type !== "week-event" &&
+            activeData.type !== "day-hourly-event" &&
+            activeData.type !== "day-list-event")
+        ) {
+          return;
+        }
+        // Clear startDate / endDate. Recurrence-aware to mirror reschedule
+        // behavior: a series occurrence opens the prompt; a one-off updates
+        // directly. startTime / endTime are explicitly cleared so the row
+        // doesn't carry a phantom time slot.
+        const patch: Partial<ProjectTask> = {
+          startDate: null,
+          endDate: null,
+          startTime: null,
+          endTime: null,
+        };
+        await dispatchUpdate(
+          calEvent.id,
+          patch,
+          "Unschedule this occurrence, or unschedule the entire series?"
+        );
+        return;
+      }
+
+      if (!overData?.day) return;
+
+      const targetDay = overData.day;
 
       // ── month-event / week-event: whole-day reschedule ───────────────
       if (
--- a/src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
+++ b/src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
@@ -24,6 +24,7 @@ import {
   useCallback,
   useContext,
   useMemo,
+  useRef,
   useState,
 } from "react";
 import {
@@ -92,6 +93,43 @@ export function useCalendarResizeContext(): CalendarResizeAPI {
   return ctx;
 }
 
+// ─── Auxiliary drag handlers ────────────────────────────────────────────────
+//
+// CrewGrid used to mount its own <DndContext> with `useCrewDnd` handlers,
+// which broke cross-surface drags (the unscheduled tray is registered with
+// the outer CalendarDndShell context, so its drags could not reach the
+// inner crew DndContext). The crew grid now registers its handler via this
+// bridge instead — CalendarDndShell.handleDragEnd invokes any registered
+// auxiliary handlers when the active drag type matches their `match` set,
+// which lets `useCrewDnd` keep its per-grid state (gridWidth ref, cascade
+// preview, dependency check) while sharing the single outer context.
+
+type AuxDragHandler = {
+  /** Set of activeData.type values this handler will receive */
+  matchTypes: Set<string>;
+  onDragStart?: (e: DragStartEvent) => void;
+  onDragEnd: (e: DragEndEvent) => void | Promise<void>;
+  onDragCancel?: (e: DragCancelEvent) => void;
+};
+
+interface CalendarDragBridgeAPI {
+  register: (id: string, handler: AuxDragHandler) => () => void;
+}
+
+const CalendarDragBridgeContext = createContext<CalendarDragBridgeAPI | null>(
+  null
+);
+
+export function useCalendarDragBridge(): CalendarDragBridgeAPI {
+  const ctx = useContext(CalendarDragBridgeContext);
+  if (!ctx) {
+    throw new Error(
+      "useCalendarDragBridge must be used inside <CalendarDndShell>"
+    );
+  }
+  return ctx;
+}
+
 // ─── Active drag descriptor ────────────────────────────────────────────────
 
 interface ActiveDrag {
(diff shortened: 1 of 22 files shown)


src/lib/hooks/use-crew-dnd.ts
```diff
@@ -267,17 +276,22 @@ export function useCrewDnd({
...
+        // on the placeholder "Unassigned" row leaves teamMemberIds empty
+        // (so the task moves out of the tray onto the unassigned strip);
+        // dropping on a member row replaces any prior assignment.
+        const newTeamMemberIds =
+          overData.teamMemberId && overData.teamMemberId !== "__unassigned__"
             ? [overData.teamMemberId]
-            : task.teamMemberIds,
+            : [];
+
+        updateMutation.mutate({
```
jack-opsapp, line 288: REVIEW PASS :: CONCERN

Recurring task scheduled silently when dropped from unscheduled tray.

If a task in the unscheduled tray belongs to a recurring series (`recurrenceId !== null` — possible when the user explicitly unscheduled an occurrence by dropping it onto the unscheduled dock, which only clears `startDate/endDate/startTime/endTime` but does not null out `recurrenceId`), this branch issues a direct `updateMutation.mutate({ id: task.id, data: {...} })` without invoking the recurrence-scope prompt (`useRecurrenceEditPrompt` / `recurrenceEdit`). The shell's other reschedule paths (`dispatchUpdate` in `calendar-dnd-shell.tsx`) correctly prompt the user with "this / following / all" before mutating recurring tasks. The aux handler bypasses that, so a series-occurrence drag from the tray onto a crew row will mutate only that one occurrence with no scope dialog — which silently diverges from the rest of the calendar's recurrence semantics and likely surprises the user.

Suggested fix: call `applyMove(task.id, patch)` (defined at line 80 of this file, which already plumbs `onRecurringEdit`) instead of `updateMutation.mutate(...)`. The hook already accepts an `onRecurringEdit` option from the caller — pass through `useRecurrenceEditPrompt` from CrewGrid the same way the existing crew-event branch does.

src/lib/hooks/use-crew-dnd.ts
```diff
@@ -267,17 +276,22 @@ export function useCrewDnd({
...
+          id: task.id,
+          data: {
+            startDate: targetDate,
+            endDate: targetEndDate,
+            teamMemberIds: newTeamMemberIds,
+          },
         });
```
jack-opsapp, line 295: REVIEW PASS :: CONCERN

No `onError` handler on the unscheduled-task → crew-row mutation. Failures are silent.

The shell's equivalent unscheduled-task branch (calendar-dnd-shell.tsx around line 465) passes `{ onError: (err) => toast.error("Failed to schedule task", { description: err.message }) }` to `updateTask.mutate`. The aux handler here does not. If the mutation fails — RLS rejection (cross-tenant `teamMemberId`), network drop mid-drag (the user may be offline; the bug log shows this is a calendar-heavy workflow), constraint violation, etc. — the optimistic cache update will be rolled back by the `useUpdateTask` `onError` snapshot path, but the user gets no toast and no indication that the drop did not persist. They will see the task snap back into the tray with no explanation.

Suggested fix: add an `onError` callback to both `updateMutation.mutate` calls in this branch (the smart-insert pushOffsets loop at line 271 and the primary mutation at line 288) that toasts the failure, matching the pattern in the shell.

src/lib/hooks/use-crew-dnd.ts
```diff
@@ -250,7 +259,7 @@ export function useCrewDnd({
 
           if (insertPoint && insertPoint.after) {
             const pushOffsets = calculatePushOffsets(
-              1, // Default 1-day duration for new tasks
+              durationDays,
```
jack-opsapp, line 262: REVIEW PASS :: CONCERN

Smart-insert push offsets bypass recurrence-scope prompt for surrounding events.

When the smart-insert path detects neighboring tasks that need to shift (the `pushOffsets` loop just below this hunk), it issues raw `updateMutation.mutate({ id: offset.eventId, data: { startDate, endDate } })` for each neighbor. If any of those neighbors are part of a recurring series, the user is never asked whether the shift should apply to that occurrence only, the following occurrences, or the whole series — the change just lands on that one occurrence. This is technically pre-existing behavior in the crew-event drag branches further down too, but the unscheduled-task → crew-row path now exercises this code through a workflow that previously did not (the old `createMutation` did not push neighbors of recurring tasks because the create path was structured differently). Worth either documenting as known limitation or routing the push offsets through `applyMove` so they participate in the recurrence prompt flow.

src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
```diff
@@ -146,35 +184,80 @@ export function CalendarDndShell({ children }: CalendarDndShellProps) {
...
+    ) => {
+      if (!activeType) return false;
+      let handled = false;
+      for (const handler of auxHandlersRef.current.values()) {
+        if (!handler.matchTypes.has(activeType)) continue;
+        const fn = handler[method];
+        if (!fn) continue;
+        // Aux handlers are typed loosely above; the runtime guarantees the
+        // event shape matches the method we're calling.
+        (fn as (evt: typeof e) => void | Promise<void>)(e);
```
jack-opsapp, line 216: REVIEW PASS :: CONCERN

`dispatchAux` invokes async aux handlers fire-and-forget — promises are not awaited.

`useCrewDnd.handleDragEnd` is `async` (it awaits `applyMove`, which awaits `onRecurringEdit`). The dispatcher in this block calls each handler synchronously and discards the returned promise. The shell's outer `handleDragEnd` is itself `async`, but it also does not `await` the aux dispatch. Two concrete consequences:

1. `setActiveDrag(null)` fires before the aux handler completes its mutation (acceptable for visual state, but means the drag overlay can disappear before the recurrence prompt renders if you ever route the crew aux handler through `applyMove`).
2. If multiple aux handlers ever register for the same type, they'll race rather than serialize.

Today there is exactly one aux handler (`crew-grid`), so the race is theoretical. But if a future surface (e.g., a second swimlane) registers another handler with overlapping `matchTypes`, the order is iteration order of a `Map`, not deterministic by intent. Worth either documenting that aux handlers must own a unique type slice (and asserting it on `register`) or making `dispatchAux` await the chain.

src/app/(dashboard)/calendar/_components/calendar-dnd-shell.tsx
```diff
@@ -183,12 +266,23 @@ export function CalendarDndShell({ children }: CalendarDndShellProps) {
             task?: { id: string; duration: number };
           }
         | undefined;
+      // Always dispatch to aux handlers first — they get the full event
+      // object (including null `over` so they can no-op cleanly) and may
+      // own the resolution for crew-event / unscheduled→crew-row /
+      // project-drawer→crew-row drops. The shell's own branches still run
+      // afterwards for any types the aux handler didn't claim (e.g. drops
+      // on month/week day cells, the unscheduled dock, etc.).
+      const auxHandled = dispatchAux("onDragEnd", dragEvent, activeData?.type);
```
jack-opsapp, line 275: REVIEW PASS :: NITPICK

`auxHandled` is set to `true` whenever any aux handler with a matching `matchTypes` runs, regardless of whether that handler actually consumed the drop.

Concretely: the crew aux handler matches `unscheduled-task` and runs even when the user drops onto a non-crew-row (e.g., they dragged from the tray, scrubbed across the crew swimlane, but released over a Week-view day cell). In that case the aux handler's body falls through (no `over.type === 'crew-row'` condition matches), so it returns without mutating — but `auxHandled` is now `true`. The shell then evaluates `if (auxHandled && overData?.type === 'crew-row') return;` — the second clause is false because the actual drop is on a Week day cell, so the shell proceeds and schedules normally. So today's logic is correct.

However, the boolean is misnamed: `auxHandled` should mean "an aux handler claimed this drop," but it actually means "an aux handler had a chance to look." A future contributor who reads the variable name and removes the second clause `&& overData?.type === 'crew-row'` would silently break Week/Month tray drops. Suggest renaming to `auxDispatched` or adding a comment that this flag is necessary-but-not-sufficient to suppress shell processing.

src/lib/hooks/use-crew-dnd.ts
```diff
@@ -235,10 +243,11 @@ export function useCrewDnd({
         // Calculate target day from horizontal drop offset
         const dayDelta = dayColumnWidth > 0 ? Math.round(delta.x / dayColumnWidth) : 0;
         const targetDate = addDays(startDate, Math.max(0, Math.min(dayDelta, daysShown - 1)));
-        const targetEndDate = addDays(targetDate, 1); // Default 1-day duration
+        const durationDays = Math.max(task.duration ?? 1, 1);
```
jack-opsapp, line 246: REVIEW PASS :: CONCERN

Default `daysShown - 1` clamp can land the dropped task on the wrong day silently if the user drops past the right edge of the visible window, or on day 0 if they drop to the left.

The `targetDate` calculation just above this line uses `Math.max(0, Math.min(dayDelta, daysShown - 1))`. If the user drops to the left of the grid (negative delta), the task lands on day 0 (`startDate`) — silently, with no indication that the drop was outside bounds. If they drop past the right edge, it lands on the last visible day. Both behaviors are silent clamps; the user just sees the task appear on the wrong date, assigned to the right person.

This is consistent with the `project-drawer-task` branch immediately below, so it's at least internally consistent. But the bug report is specifically about "schedule for the date on which I dropped" — a silent clamp may surface as a separate user-reported bug ("I dragged into Friday but it landed on Monday"). Combined with the missing `onError` toast (other comment), there is no feedback at all. Suggest either rejecting out-of-range drops with a toast, or visually constraining the drop preview so the user can't release outside bounds.

jack-opsapp:
REVIEW PASS (bug 1b2942d5) :: 6 concerns (0 blockers, 5 concerns, 1 nitpick)

STATUS :: ALLOW MERGE

The fix correctly addresses the reported bug: the nested `<DndContext>` removal and switch from `createMutation` to `updateMutation` resolve both the "drag from unscheduled tray to crew row does nothing" and the "duplicate task created" failure modes. The bridge pattern (`CalendarDragBridgeContext`) is sound and the cleanup contract is correct.

No blockers. The concerns are quality/edge-case items that should be tracked but do not gate this merge:

1. **Recurrence-scope prompt skipped** when an unscheduled tray task with `recurrenceId !== null` is dropped onto a crew row. The shell uses `applyMove` / `dispatchUpdate` to prompt for series scope; the aux handler bypasses both. (use-crew-dnd.ts L288)
2. **No `onError` toast** on the unscheduled-task → crew-row mutation. Other shell branches toast on failure; this one fails silently. (use-crew-dnd.ts L295)
3. **Push-offset shifts on neighbors** of the dropped task also bypass the recurrence prompt — pre-existing in crew-event drag, but newly exercised by this code path. (use-crew-dnd.ts L262)
4. **Async aux handlers fire-and-forget** — `dispatchAux` does not await returned promises. Today only one handler is registered, so the race is theoretical, but worth either documenting or awaiting. (calendar-dnd-shell.tsx L216)
5. **Day-clamp is silent** — drops to the left of the grid land on day 0; drops past the right edge land on the last day. Combined with the missing `onError` toast, the user gets no feedback that the drop landed in the wrong place. (use-crew-dnd.ts L246)
6. **`auxHandled` boolean misnamed** — actually means "an aux handler had a chance to look," not "an aux handler claimed the drop." A future contributor could misread it. (calendar-dnd-shell.tsx L275, NITPICK)

PR #32, opened 2026-04-30
Nightly Bug Sweep — web — 2026-04-30 — Batch 03

jack-opsapp:
Final batch (3 of 3) of the nightly web bug sweep for 2026-04-30. All three bugs fixed; one is a high-priority crash recovery for the inbox route.

### a8ece79d — bug — urgent — Inbox crashed with `Cannot read properties of undefined (reading 'label')` (86 errors) `[COMPLEX]`
The `CategoryChip` lookup `CATEGORY_STYLES[category].label` blew up whenever a thread row arrived with a `primaryCategory` outside the hardcoded enum (legacy rows, server-side enum extension, or malformed payload). Hardened the read site in `src/components/ops/inbox/category-chip.tsx` with a `resolveCategoryStyle()` helper that falls back to OTHER's style, and a `categoryLabel()` helper that surfaces the raw token (uppercased) when present so the operator can still see what came back. The whole inbox route no longer hard-stops on a single unmapped category.

S3 CORS errors on `ops-app-files-prod.s3.us-west-2.amazonaws.com` for profile avatars are a separate concern (S3 bucket policy, not in this repo) — noted in `fix_notes` but not fixed here.

### dd5659ed — ui_issue — high — Quick Actions tab much too tall
The Quick Actions tab is paired with a panel-anchored drawer (308×452), not a full-rail drawer. EdgeTab's legacy expansion (`top:0; height:100%`) made the tab much taller than its menu. Passed `expandedHeight={452}` (matching `PANEL_H` in `quick-actions-drawer.tsx`) so the tab now grows only as tall as the panel. Builds on the EdgeTab overhaul below.

### 85da1e52 — ui_issue — high — Hovering one tab covered the other
The right-edge action tabs (Notifications + Quick Actions) shared a single expand path that set `top:0; height:100%` on hover OR open. On hover alone, the hovered tab burst across the rail midpoint and covered the sibling. Split the behavior in `src/components/ui/edge-tab.tsx`:
- Open with `expandedHeight` → centered on `stackOffset`, height clamped to `expandedHeight` (matches the panel-anchored drawer geometry).
- Open without `expandedHeight` → legacy full-rail (Notifications continues to use this).
- Hover with `expandedHeight` → grow outward from the near edge, capped at `expandedHeight`, never crossing the rail midpoint.
- Hover without `expandedHeight` → stay at rest height, rather than bursting full-rail.

## Test plan

- [ ] Open `/inbox` on a company with threads carrying any (or no) `primaryCategory` and verify the page renders instead of crashing with the route-error fallback.
- [ ] Force a thread to have an unmapped category (e.g. via DB) and confirm the chip shows the raw token uppercased and the rest of the inbox works.
- [ ] Open `/calendar` (or any page where the Quick Actions tab renders) and hover the QA tab — verify it grows only to the height of the menu (452px), not the whole rail.
- [ ] Hover the Notifications tab — it stays at rest height (180px) rather than going full rail; the QA tab below remains visible.
- [ ] Click to open the QA drawer — the tab + drawer share the same vertical span (centered around `+94` from rail midpoint, 452px tall).
- [ ] Click to open the Notifications drawer — full-rail expansion still works as before.
- [ ] `npm run type-check && npm run build` passes.

### DIFF
--- a/src/components/ops/inbox/category-chip.tsx
+++ b/src/components/ops/inbox/category-chip.tsx
@@ -41,12 +41,32 @@ const CATEGORY_STYLES: Record<EmailThreadCategory, CategoryStyle> = {
   OTHER:        { border: "#4a4a4a", label: "OTHER",        dotColor: "#4a4a4a" },
 };
 
+// Fallback used when a thread arrives with a `primaryCategory` that isn't
+// in CATEGORY_STYLES — e.g. a legacy row, a server-side enum extension, or
+// a malformed payload. We render OTHER's style and the raw value (if any)
+// rather than crashing the whole inbox route. (See bug a8ece79d.)
+const FALLBACK_STYLE: CategoryStyle = CATEGORY_STYLES.OTHER;
+
+function resolveCategoryStyle(
+  category: EmailThreadCategory | null | undefined
+): CategoryStyle {
+  if (!category) return FALLBACK_STYLE;
+  return CATEGORY_STYLES[category] ?? FALLBACK_STYLE;
+}
+
 export function categoryLabel(category: EmailThreadCategory): string {
-  return CATEGORY_STYLES[category].label;
+  const style = CATEGORY_STYLES[category];
+  if (style) return style.label;
+  // For unmapped categories, surface the raw token (uppercased) so the
+  // operator can still see what came back from the server.
+  if (typeof category === "string" && category.length > 0) {
+    return category.replace(/_/g, " ").toUpperCase();
+  }
+  return FALLBACK_STYLE.label;
 }
 
 export function categoryDotColor(category: EmailThreadCategory): string {
-  return CATEGORY_STYLES[category].dotColor;
+  return (CATEGORY_STYLES[category] ?? FALLBACK_STYLE).dotColor;
 }
 
 // ─── Chip props ──────────────────────────────────────────────────────────────
@@ -78,8 +98,8 @@ export const CategoryChip = forwardRef<HTMLButtonElement | HTMLSpanElement, Cate
     { category, size = "sm", interactive = false, onClick, label, leading, manual, className, title },
     ref
   ) {
-    const style = CATEGORY_STYLES[category];
-    const display = label ?? style.label;
+    const style = resolveCategoryStyle(category);
+    const display = label ?? categoryLabel(category);
 
     const sizeClasses =
       size === "md"
--- a/src/components/layouts/quick-actions-tab.tsx
+++ b/src/components/layouts/quick-actions-tab.tsx
@@ -12,6 +12,11 @@ const EDGE_TAB_ID = "quick-actions";
 const STACK_OFFSET_QA = 94;
 const REST_HEIGHT = 132;
 const DRAWER_WIDTH = 308;
+// Match the panel-anchored drawer (`PANEL_H`) in quick-actions-drawer.tsx so
+// the tab + panel read as one shape when expanded, and the tab never towers
+// above its drawer or overlaps the Notifications tab. (See bugs dd5659ed +
+// 85da1e52.)
+const EXPANDED_HEIGHT_QA = 452;
 
 export function QuickActionsTab() {
   const { t } = useDictionary("quick-actions");
@@ -44,6 +49,7 @@ export function QuickActionsTab() {
       onToggle={() => toggle(EDGE_TAB_ID)}
       accent="accent"
       restHeight={REST_HEIGHT}
+      expandedHeight={EXPANDED_HEIGHT_QA}
       drawerWidth={DRAWER_WIDTH}
       stackOffset={STACK_OFFSET_QA}
       canHoverExpand={!anyActive || open}
--- a/src/components/ui/edge-tab.tsx
+++ b/src/components/ui/edge-tab.tsx
@@ -25,6 +25,7 @@ export function EdgeTab({
   count,
   accent = "accent",
   restHeight = DEFAULT_REST_HEIGHT,
+  expandedHeight,
   drawerWidth = DEFAULT_DRAWER_WIDTH,
   railTop = DEFAULT_RAIL_TOP,
   railBottom = DEFAULT_RAIL_BOTTOM,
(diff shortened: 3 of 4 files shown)


src/components/ops/inbox/category-chip.tsx
```diff
@@ -41,12 +41,32 @@ const CATEGORY_STYLES: Record<EmailThreadCategory, CategoryStyle> = {
...
+const FALLBACK_STYLE: CategoryStyle = CATEGORY_STYLES.OTHER;
+
+function resolveCategoryStyle(
+  category: EmailThreadCategory | null | undefined
+): CategoryStyle {
+  if (!category) return FALLBACK_STYLE;
+  return CATEGORY_STYLES[category] ?? FALLBACK_STYLE;
+}
+
 export function categoryLabel(category: EmailThreadCategory): string {
```
jack-opsapp, line 57: REVIEW PASS :: CONCERN

The public type signatures of `categoryLabel(category: EmailThreadCategory)` and `categoryDotColor(category: EmailThreadCategory)` still claim to accept ONLY a valid `EmailThreadCategory`, but the new bodies defensively handle `null`, `undefined`, and unknown string tokens. This is a lie to TypeScript: callers like `src/components/ops/inbox/category-filter-chips.tsx:153-154` already cast `value as EmailThreadCategory` from a `value` that can be `null` — TS lets it slide because the cast is unchecked. Suggested fix: widen both signatures to `EmailThreadCategory | null | undefined` (or `string | null | undefined`) so the runtime guard is reflected in the type system and callers don't need unsafe casts. Otherwise the next caller will pass `undefined` and get a type-system green light without realizing they're depending on a runtime fallback.

src/components/ops/inbox/category-chip.tsx
```diff
@@ -78,8 +98,8 @@ export const CategoryChip = forwardRef<HTMLButtonElement | HTMLSpanElement, Cate
     { category, size = "sm", interactive = false, onClick, label, leading, manual, className, title },
```
jack-opsapp, line 98: REVIEW PASS :: CONCERN

The `CategoryChip` props still type `category: EmailThreadCategory` (non-nullable). The fix hardens the runtime, but the *root cause* is upstream — somewhere a thread is arriving with a `primaryCategory` that violates the declared `EmailThreadCategory` enum (legacy row, server enum extension, or the API caster at `src/lib/types/email-thread.ts:223` doing `row.primary_category as EmailThreadCategory` without validating). The chip patch is fine as a guardrail, but consider also: (a) validating `primary_category` at the type-cast boundary in `email-thread.ts` so unknown values get coerced to `OTHER` once, centrally, before they propagate; or (b) widening `CategoryChipProps.category` to `EmailThreadCategory | string | null | undefined` so this defensive contract is visible at every call site (e.g. `conversation-list.tsx:311`, `thread-context-panel.tsx:292`, `thread-sibling-strip.tsx:241`). Otherwise we're permanently shipping a chip whose props lie about what they accept.

src/components/ops/inbox/category-chip.tsx
```diff
@@ -78,8 +98,8 @@ export const CategoryChip = forwardRef<HTMLButtonElement | HTMLSpanElement, Cate
...
     ref
   ) {
-    const style = CATEGORY_STYLES[category];
-    const display = label ?? style.label;
+    const style = resolveCategoryStyle(category);
```
jack-opsapp, line 101: REVIEW PASS :: NITPICK

When `category` is `null`/`undefined` and the chip falls back to OTHER, the rendered button still emits `data-category={category}` (lines 152, 165 of the file — outside this diff hunk) which becomes `data-category=""` or omits the attribute entirely. If anything in the codebase queries by `[data-category="OTHER"]` (e.g. tests, e2e selectors, analytics) it will silently miss these fallback chips. Consider also propagating the resolved style here: `data-category={style === FALLBACK_STYLE ? "OTHER" : category}` so the DOM stays consistent with the rendered visual. Low impact, easy fix.

src/components/ops/inbox/category-chip.tsx
```diff
@@ -41,12 +41,32 @@ const CATEGORY_STYLES: Record<EmailThreadCategory, CategoryStyle> = {
...
+// in CATEGORY_STYLES — e.g. a legacy row, a server-side enum extension, or
+// a malformed payload. We render OTHER's style and the raw value (if any)
+// rather than crashing the whole inbox route. (See bug a8ece79d.)
+const FALLBACK_STYLE: CategoryStyle = CATEGORY_STYLES.OTHER;
+
+function resolveCategoryStyle(
+  category: EmailThreadCategory | null | undefined
+): CategoryStyle {
+  if (!category) return FALLBACK_STYLE;
+  return CATEGORY_STYLES[category] ?? FALLBACK_STYLE;
```
jack-opsapp, line 54: REVIEW PASS :: CONCERN

If `thread.primaryCategory` arrives as an unknown token (the very case this PR guards against), the user sees a chip showing the raw token (uppercased). When they open the recategorize menu and pick a real category, then click UNDO on the toast, `recategorize-menu.tsx:97-101` calls `recategorize.mutate({ toCategory: currentCategory })` — sending the same invalid category string back to the server. Depending on the server's validation, that either: (a) silently fails, (b) writes the invalid token back, or (c) 400s and shows an error to the user. Not a crash, but a real bug surfaced by exactly the data shape this PR exists to handle. Suggested fix: in `recategorize-menu.tsx`, if `currentCategory` is not in `EMAIL_THREAD_CATEGORIES`, disable the undo (or set the undo target to `OTHER`). Out of scope for an urgent crash fix, but worth a follow-up issue.

src/components/ops/inbox/category-chip.tsx
```diff
@@ -41,12 +41,32 @@ const CATEGORY_STYLES: Record<EmailThreadCategory, CategoryStyle> = {
   OTHER:        { border: "#4a4a4a", label: "OTHER",        dotColor: "#4a4a4a" },
 };
 
+// Fallback used when a thread arrives with a `primaryCategory` that isn't
+// in CATEGORY_STYLES — e.g. a legacy row, a server-side enum extension, or
+// a malformed payload. We render OTHER's style and the raw value (if any)
+// rather than crashing the whole inbox route. (See bug a8ece79d.)
+const FALLBACK_STYLE: CategoryStyle = CATEGORY_STYLES.OTHER;
```
jack-opsapp, line 48: REVIEW PASS :: CONCERN

The console errors include S3 CORS failures on profile avatars: `Access to fetch at 'https://ops-app-files-prod.s3.us-west-2.amazonaws.com/.../profile_....jpg' from origin 'https://app.opsapp.co' has been blocked by CORS policy`. The fix-notes mark this as "out of scope, S3 bucket policy." That's correct for the *root* cause (S3 needs `Access-Control-Allow-Origin` configured), but note that the message specifically says *fetch* — meaning some code path is using `fetch()` to retrieve the avatar (or an `` attribute). Plain `` is NOT CORS-blocked from rendering. Two frontend mitigations worth considering as a follow-up:

1. If the code is using `fetch` to prefetch/cache avatars, drop the fetch and let the browser load the `` natively (no CORS check unless `crossOrigin` is set).
2. Add an `onError` handler on the avatar `` that swaps to a default avatar so users never see broken images while the bucket policy is being fixed server-side.

Not blocking this PR, but the fix-notes shouldn't imply there's nothing the frontend can do.

jack-opsapp:
REVIEW PASS (bug a8ece79d) :: 5 concerns (0 blockers, 4 concerns, 1 nitpick)

STATUS :: ALLOW MERGE

The crash fix itself is correct and sufficient to stop the inbox route from failing — `resolveCategoryStyle()` and the hardened `categoryLabel`/`categoryDotColor` handle null, undefined, and unknown string tokens. No blockers.

Follow-ups (non-blocking):
1. **Type signatures lie about runtime behavior** — `categoryLabel` and `categoryDotColor` still type `category: EmailThreadCategory` despite the body now accepting null/undefined. Widen the signatures so callers don't need unsafe casts.
2. **Root cause not addressed** — the chip now tolerates bad data, but the API caster at `src/lib/types/email-thread.ts:223` (`row.primary_category as EmailThreadCategory`) is the source of the violation. Validate at that boundary so the rest of the codebase can trust the type.
3. **`CategoryChipProps.category` still non-nullable** — same shape problem, callers like `conversation-list.tsx:311` won't get TS warnings if they pass invalid data.
4. **Undo path can re-send invalid token** — if `currentCategory` is the unknown token this PR exists to handle, the recategorize-menu undo will send it back to the server.
5. **CORS on avatars marked out-of-scope, but frontend mitigations exist** — error says *fetch*-blocked, suggesting code is calling `fetch()` for avatars (or using `crossOrigin="anonymous"`). Worth checking; also add an `onError` fallback to a default avatar so users don't see broken images while the bucket policy gets fixed server-side.

PR #39, opened 2026-05-08
feat(products): FK-backed category + unit pickers with inline create

jack-opsapp:
## Summary

- Replaces the free-text category Input and the hardcoded UNIT_OPTIONS select on the products form with FK-backed pickers (Combobox over `catalog_categories` and `catalog_units`) plus inline-create dialogs that mirror the iOS `InlineCreateCategorySheet` / `InlineCreateUnitSheet` UX.
- Web now writes `category_id` and `unit_id` directly via the picker callbacks; the legacy `category` / `unit` text columns stay in lockstep for backward compat.
- Adds `CatalogCategoryService.create` + `CatalogUnitService.create` (auto-computes next `sort_order`, mirrors iOS DTO field set) and `useCreateCatalogCategory` / `useCreateCatalogUnit` mutation hooks that invalidate the lookup cache on success.

**Stacks on #38** (FK writes). Merge order: #38 first, then this.

## Verification (run after deploy)

```sql
-- New products via web after deploy: confirm both FK + legacy text written
SELECT id, name, category, category_id, unit, unit_id, created_at
FROM products
WHERE deleted_at IS NULL AND created_at > now() - interval '24 hours'
ORDER BY created_at DESC;
```

## Test plan

- [ ] Open Products → Add Item → category picker shows existing categories with search, unit picker shows units alphabetically with abbreviations
- [ ] Pick a category → form submits with `category_id` populated and `category` text mirrored
- [ ] Pick "+ NEW CATEGORY…" → dialog opens with name field autofocused → save → new row inserted, picker selects it automatically
- [ ] Pick "+ NEW UNIT…" → dimension defaults to "count", radio group switches dimension, save → new row inserted, picker selects it
- [ ] Edit a legacy product (no FK ids, only legacy text) → backfill matches the text and picker shows the matching row
- [ ] Edit a product with FK ids → pickers immediately reflect the FK
- [ ] iOS: open the same product on iOS → category + unit display matches what web wrote

## Notes

- Dialogs use the shared `Dialog`/`DialogContent` primitive which already applies `glass-dense`. No divergence from spec.
- Pickers' popovers stack above the form modal via Radix portal (z-index `modal: 3000`). Verified visually consistent.
- 11 new unit tests across both pickers; full suite passes (897 / 5 skipped).

### DIFF
--- a/src/app/(dashboard)/products/page.tsx
+++ b/src/app/(dashboard)/products/page.tsx
@@ -20,6 +20,10 @@ import {
 import { Textarea } from "@/components/ui/textarea";
 import { EmptyState } from "@/components/ops/empty-state";
 import { ProductBomEditor } from "@/components/ops/product-bom-editor";
+import { CategoryPicker } from "@/components/ops/category-picker";
+import { UnitPicker } from "@/components/ops/unit-picker";
+import { InlineCreateCategoryDialog } from "@/components/ops/inline-create-category-dialog";
+import { InlineCreateUnitDialog } from "@/components/ops/inline-create-unit-dialog";
 import {
   useProducts,
   useCreateProduct,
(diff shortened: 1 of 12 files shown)


jack-opsapp:
Closing — base branch `feat/products-fk-writes` no longer exists. Reopen against `main` if this work is still needed.

PR #52, opened 2026-05-11
Nightly Bug Sweep — web — 2026-05-11 — Batch 01

jack-opsapp:
## Summary

Batch 01 of the 2026-05-11 web nightly bug sweep — 8 fixed, 2 escalated.

## Fixes

- **7ea0db1b** · security · Protect `/inbox` and `/agent` prefixes in middleware. Added both routes to `protectedPrefixes` so unauthenticated requests are redirected at the edge before client hydration.
- **6c4f2a6c** · ui · Dictionary-back FAB quick action labels. Added `labelKey` to FABAction, populated `quick-actions.json` for EN + ES, and updated drawer + settings tabs to resolve via the dictionary using a `resolveActionLabel` helper.
- **63d0cf8d** · functional · Unify Agent Queue permission. Introduced `agent.queue.view` permission, used it in both sidebar nav and the dashboard route gate. Resolves the prior mismatch where the sidebar's "admin" key was a category id (not a permission) and the route mapped /agent to pipeline.view.
- **c0d4abe7** · functional · **COMPLEX** · Persist schedule edits from project task list edit form. `handleEditSubmit` now converts the form's `YYYY-MM-DD` strings to `Date | null` and includes `startDate` + `endDate` in the partial update so `TaskService.updateTask` writes `start_date`/`end_date` columns.
- **29fcbf09** · ui · **COMPLEX** · Show task assignee, schedule, and overflow controls on mobile. Removed `hidden sm:flex` from the assignee + date-range groups and dropped the hover-only opacity on the overflow trigger. Added an aria-label.
- **452d7865** · functional · **COMPLEX** · Hide In Progress from project task status menus. The `project_tasks.status` CHECK constraint only allows active/completed/cancelled — the UI exposed In Progress, but writes collapsed to active and reads mapped back to Booked. Removed from `STATUS_OPTIONS` and filtered out of `StatusDropdown` until a future migration widens the constraint.
- **03241853** · ui · Added shared `QueryErrorState` component (brick-bordered tactical panel + retry button with spinner) and wired it into calendar, projects, map, and team pages. Error pane only swaps in when no cached data is present; otherwise the prior result remains visible while the background refetch retries.
- **272b3751** · functional · Wired project Record Payment action to the invoices flow. The action now routes to `/invoices?projectId={id}&action=recordPayment`, the invoices page filters by project and auto-opens the existing Record Payment modal when there's one outstanding invoice (toasts a dictionary-backed "no outstanding invoices" message otherwise). Re-gated `canRecordPayment` on the project canvas to `invoices.record_payment` to match the destination.

## Escalated (requires human review)

- **ae2cf28a** · functional · Product options TypeScript build failure. None of the referenced files exist on main and `npm run type-check` passes cleanly. Likely filed against an unmerged branch — needs a human to confirm whether to scaffold the missing module or close as obsolete.
- **fcd58fbb** · functional · Product options page JSX comment lint. Same surface as `ae2cf28a` — `src/app/(dashboard)/products/[id]/options/page.tsx` does not exist on main. Cannot run/repair lint when no source file exists.

## Test plan

- [ ] Unauthenticated `/inbox` and `/agent/queue` requests redirect to `/login` (edge, not client).
- [ ] FAB quick action drawer labels render dictionary copy in EN and ES; settings tabs render the same copy.
- [ ] Sidebar shows Agent Queue entry for account holders / company admins; route gate 404s non-admins.
- [ ] Editing a task's start/end date in the project task list persists across refresh.
- [ ] Task row on mobile (≤ 640px) shows assignee avatars, date range, and an always-visible overflow trigger; tapping the trigger opens Edit/Delete.
- [ ] Setting task status — In Progress no longer appears in the inline menu or the form dropdown.
- [ ] Forcing a network failure on calendar / projects / map / team shows the QueryErrorState pane with a working Retry button.
- [ ] Clicking Record Payment on a project with one outstanding invoice opens the Record Payment modal on the invoices page; with zero outstanding invoices it shows the dictionary-backed toast.

### DIFF
--- a/src/app/(dashboard)/invoices/page.tsx
+++ b/src/app/(dashboard)/invoices/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useMemo, useEffect, useCallback } from "react";
+import { useState, useMemo, useEffect, useCallback, useRef } from "react";
 import { useSearchParams } from "next/navigation";
 import { usePageTitle } from "@/lib/hooks/use-page-title";
 import { useDictionary, useLocale } from "@/i18n/client";
@@ -206,8 +206,17 @@ export default function InvoicesPage() {
     return map;
   }, [projects]);
 
+  // Project filter pulled from the URL — set when the projects canvas
+  // hands off to the invoices page for a "Record payment" workflow
+  // (bug 272b3751). Kept reactive on every query-string change so the
+  // user can clear it via the back button without an extra re-route.
+  const projectIdFilter = searchParams.get("projectId");
+
   const filtered = useMemo(() => {
     let list = [...invoices];
+    if (projectIdFilter) {
+      list = list.filter((i) => i.projectId === projectIdFilter);
+    }
     if (filterStatus !== "all") {
       list = list.filter((i) => i.status === filterStatus);
     }
@@ -220,7 +229,34 @@ export default function InvoicesPage() {
       );
     }
     return list;
-  }, [invoices, filterStatus, searchQuery, clientMap]);
+  }, [invoices, projectIdFilter, filterStatus, searchQuery, clientMap]);
+
+  // Auto-open the record-payment modal when the project canvas routes
+  // here with action=recordPayment AND there's exactly one outstanding
+  // invoice for the project — the common case. If there's more than
+  // one, we leave the user on the filtered list to choose.
+  const autoPaymentRanRef = useRef(false);
+  useEffect(() => {
+    if (autoPaymentRanRef.current) return;
+    if (searchParams.get("action") !== "recordPayment") return;
+    if (!projectIdFilter) return;
+    if (!can("invoices.record_payment")) return;
+    const outstanding = invoices.filter(
+      (i) =>
+        i.projectId === projectIdFilter &&
+        i.status !== InvoiceStatus.Paid &&
+        i.status !== InvoiceStatus.Void &&
+        i.status !== InvoiceStatus.Draft,
+    );
+    if (outstanding.length === 1) {
+      setPaymentInvoice(outstanding[0]);
+      autoPaymentRanRef.current = true;
+    } else if (outstanding.length === 0 && invoices.length > 0) {
+      // No outstanding invoices to apply payment to — tell the user.
+      toast.info(t("invoices.payment.noOutstanding"));
+      autoPaymentRanRef.current = true;
+    }
+  }, [searchParams, projectIdFilter, invoices, can, t]);
 
   const metrics = useMemo(() => {
     // Drafts are not yet sent and therefore not "outstanding" — exclude them
--- a/src/app/(dashboard)/projects/page.tsx
+++ b/src/app/(dashboard)/projects/page.tsx
@@ -3,6 +3,7 @@
 import { useState, useCallback, useMemo, useEffect, useRef, memo } from "react";
 import { Loader2 } from "lucide-react";
 import { useDictionary } from "@/i18n/client";
+import { QueryErrorState } from "@/components/ops/query-error-state";
 import { usePageTitle } from "@/lib/hooks/use-page-title";
 import { trackScreenView } from "@/lib/analytics/analytics";
 import { toast } from "@/components/ui/toast";
@@ -185,7 +186,13 @@ export default function ProjectsPage() {
   const canDelete = can("projects.delete");
 
   // ── Data fetching ──
-  const { data: projectsData, isLoading } = useScopedProjects();
+  const {
+    data: projectsData,
+    isLoading,
+    isError: projectsError,
+    isFetching: projectsFetching,
+    refetch: refetchProjects,
+  } = useScopedProjects();
   const { data: clientsData } = useClients();
   const { data: teamData } = useTeamMembers();
   const { data: invoicesData } = useInvoices();
@@ -682,6 +689,25 @@ export default function ProjectsPage() {
   const batchCount = activeCardId && useProjectCanvasStore.getState().selectedCardIds.has(activeCardId)
     ? selectedCount
     : 1;
+
+  // ── Error ──
+  // Surface a retry path on full failure so the canvas isn't stuck on
+  // the spinner state (bug 03241853). We only swap in the error pane
+  // when there's no cached data — if we already have a previous result
+  // the canvas renders that and the background refetch continues silently.
+  if (projectsError && !projectsData) {
+    return (
+      <div className="flex flex-col h-full items-center justify-center gap-3 p-4">
+        <QueryErrorState
+          title="Could not load projects."
+          description="The project canvas couldn't reach the server. Try again — your unsaved layout edits are preserved locally."
+          errorCode="PROJECTS_CANVAS"
+          onRetry={() => refetchProjects()}
+          isRetrying={projectsFetching}
+        />
+      </div>
+    );
+  }
 
   // ── Loading ──
   if (isLoading) {
--- a/src/app/(dashboard)/projects/page.tsx
+++ b/src/app/(dashboard)/projects/page.tsx
@@ -2,6 +2,7 @@
 
 import { useState, useCallback, useMemo, useEffect, useRef, memo } from "react";
 import { Loader2 } from "lucide-react";
+import { useRouter } from "next/navigation";
 import { useDictionary } from "@/i18n/client";
 import { QueryErrorState } from "@/components/ops/query-error-state";
 import { usePageTitle } from "@/lib/hooks/use-page-title";
@@ -175,6 +176,7 @@ export default function ProjectsPage() {
   usePageTitle("Projects");
   const { t } = useDictionary("projects-canvas");
   const { can } = usePermissionStore();
+  const router = useRouter();
   const { isComplete: setupComplete, needsWebSetup, missingSteps } = useSetupGate();
   const [showSetupModal, setShowSetupModal] = useState(false);
 
@@ -182,7 +184,10 @@ export default function ProjectsPage() {
   const canManage = can("projects.edit");
   const canViewAccounting = can("accounting.view");
   const canCreateTasks = can("tasks.create");
-  const canRecordPayment = can("accounting.edit");
+  // Record-payment lives on the invoices surface — gate the project
+  // canvas action on the same permission the destination page enforces
+  // so we never advertise an action a user can't complete (bug 272b3751).
+  const canRecordPayment = can("invoices.record_payment");
   const canDelete = can("projects.delete");
 
   // ── Data fetching ──
@@ -563,10 +568,20 @@ export default function ProjectsPage() {
     [openWindow, projectMap]
   );
 
-  const handleRecordPayment = useCallback((_projectId: string) => {
-    // TODO: Open payment recording form
-    toast.info("Payment recording coming soon");
-  }, []);
+  // Hand off to the invoices page filtered to this project with the
+  // record-payment intent. The invoices page auto-opens the payment
+  // modal when there's exactly one outstanding invoice and toasts a
+  // clear message when there are none — that's a real workflow rather
+  // than the previous "Payment recording coming soon" dead end
+  // (bug 272b3751).
+  const handleRecordPayment = useCallback(
+    (projectId: string) => {
+      router.push(
+        `/invoices?projectId=${encodeURIComponent(projectId)}&action=recordPayment`,
+      );
+    },
+    [router],
+  );
 
   const handleArchive = useCallback(
     (projectId: string) => {
--- a/src/components/ops/task-list.tsx
+++ b/src/components/ops/task-list.tsx
@@ -97,9 +97,18 @@ function taskStatusToKey(status: TaskStatus): StatusBadgeTaskStatus {
   }
 }
 
+// Project task statuses available in the inline status menu.
+// project_tasks.status has a CHECK constraint of ('active','completed','cancelled')
+// — In Progress is intentionally NOT a persistable slot at the DB layer
+// (the migrate_task_status_to_active migration collapsed Booked+InProgress
+// into a single 'active' state, see task-service.ts:serializeTaskStatus).
+// Exposing it in the menu caused users to "set" In Progress only to have
+// the value collapse to active on write and read back as Booked on the
+// next refetch (bug 452d7865). The TS enum keeps the value for iOS
+// parity but no surface should offer it for project tasks until the
+// constraint is widened in a future migration.
 const STATUS_OPTIONS: { value: TaskStatus; label: string }[] = [
   { value: TaskStatus.Booked, label: "Booked" },
-  { value: TaskStatus.InProgress, label: "In Progress" },
   { value: TaskStatus.Completed, label: "Completed" },
   { value: TaskStatus.Cancelled, label: "Cancelled" },
 ];
--- a/src/components/ops/task-list.tsx
+++ b/src/components/ops/task-list.tsx
@@ -499,6 +499,16 @@ function TaskList({ projectId, companyId, className }: TaskListProps) {
     (values: TaskFormValues) => {
       if (!editingTask) return;
 
+      // Schedule fields come from the form as `YYYY-MM-DD` strings (or "").
+      // Convert to Date | null so the update mirrors the shape of
+      // ProjectTask.startDate/endDate. Sending undefined here would let the
+      // partial update skip these fields entirely — the bug (c0d4abe7) was
+      // dispatchers editing the schedule and the new dates never persisting.
+      const startDate = values.startDate
+        ? new Date(values.startDate)
+        : null;
+      const endDate = values.endDate ? new Date(values.endDate) : null;
+
       updateTaskMutation.mutate(
         {
           id: editingTask.id,
@@ -507,6 +517,8 @@ function TaskList({ projectId, companyId, className }: TaskListProps) {
             taskTypeId: values.taskTypeId,
             taskColor: values.taskColor || "#59779F",
             teamMemberIds: values.teamMemberIds || [],
+            startDate,
+            endDate,
           },
         },
         {
--- a/src/components/ops/task-list.tsx
+++ b/src/components/ops/task-list.tsx
@@ -208,8 +208,11 @@ function TaskRow({
         {title}
       </p>
 
-      {/* Assignee avatars */}
-      <div className="hidden sm:flex items-center shrink-0">
+      {/* Assignee avatars — visible at every breakpoint so phone users
+          can re-assign or open the team picker directly from the row.
+          Was previously `hidden sm:flex`, which hid the control on mobile
+          and made the row a dead end for field crews (bug 29fcbf09). */}
+      <div className="flex items-center shrink-0">
         {assignedMembers.length > 0 ? (
           <div className="flex items-center">
             {assignedMembers.slice(0, 3).map((member, idx) => (
@@ -244,8 +247,11 @@ function TaskRow({
         )}
       </div>
 
-      {/* Date range */}
-      <div className="hidden sm:flex items-center shrink-0">
+      {/* Date range — visible at every breakpoint so the unscheduled
+          badge / mini-calendar popover stays reachable on mobile. The
+          start/end date string itself is short (e.g. "Jul 14") and
+          collapses gracefully alongside the truncating task title. */}
+      <div className="flex items-center shrink-0">
         {hasStartDate ? (
           <span
             className={cn(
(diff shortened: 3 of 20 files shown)


src/lib/types/permissions.ts
```diff
@@ -301,6 +301,21 @@ const reportsModule: PermissionModule = {
...
+// /api/agent/queue uses requireAdminOrOwner — only account holders and
+// company admins should see the nav entry or reach the page. Adding the
+// permission here also keeps the route gate, sidebar, and ALL_PERMISSIONS
+// lookup consistent (bug 63d0cf8d — previously the sidebar used the
+// non-existent "admin" key and the route gate fell back to pipeline.view).
+const agentModule: PermissionModule = {
+  id: "agent",
+  label: "Agent",
+  actions: [
+    { id: "agent.queue.view", label: "View agent action queue", scopes: ["all"] },
```
jack-opsapp, line 315: REVIEW PASS :: BLOCKER

The new `agent.queue.view` permission is granted by `usePermissionStore` to anyone where `company.accountHolderId === userId` OR `userId` is in `company.adminIds` (see src/lib/store/permissions-store.ts:79-97). But the API at /api/agent/queue still uses `requireAdminOrOwner` (src/app/api/agent/_lib/auth.ts:46), which only accepts users whose `users.role` column is literally `"admin"` or `"owner"`.

These two checks are independent — `company.adminIds` membership is NOT the same as `users.role = "admin"`. So a user listed in `adminIds` but with `role = "office"` (or anything else) now:

1. Sees the sidebar entry (route gate grants `agent.queue.view`)
2. Reaches the page (route gate passes)
3. Gets 403 on every API call (`useApprovalQueue`, `useApprovalQueueStats`, mutations)
4. Page renders empty / errored

The commit message claims "non-admins now get the same 404 the API already enforces" — but the page gate and API gate use different definitions of "admin." The unification is incomplete.

Suggested fix: either (a) align `requireAdminOrOwner` to also accept `company.adminIds`/`accountHolderId` membership, or (b) align the permission store to grant `agent.queue.view` only when `users.role` is admin/owner. Option (a) matches the rest of the permission system.

src/components/ops/task-list.tsx
```diff
@@ -507,6 +517,8 @@ function TaskList({ projectId, companyId, className }: TaskListProps) {
             taskTypeId: values.taskTypeId,
             taskColor: values.taskColor || "#59779F",
             teamMemberIds: values.teamMemberIds || [],
+            startDate,
+            endDate,
```
jack-opsapp, line 541: Spurious reschedule emails: the form input is YYYY-MM-DD only, but the prior task.start_date in DB carries a real time-of-day (e.g. `2026-05-11T13:00:00Z`). On submit, `new Date(values.startDate)` produces UTC midnight (`2026-05-11T00:00:00Z`). In task-service.ts:493-497 the change detection compares `priorStartIso !== newStart.toISOString()` — these will almost never match, so `startChanged` will be true on every edit submit for any task whose start was set with a non-midnight time. Combined with `wasConfirmed`, this fires `onConfirmedTaskRescheduled`. If the company has `appointment_confirmation.reschedule_behavior` set to `draft` or `auto_send`, a customer-facing reschedule email is drafted/sent every time a dispatcher opens this form and saves — even if they only changed status, type, or team members. Before this fix `data.startDate` was always undefined, so the hook never fired from this surface. This is a regression. Fix options: (a) only send startDate/endDate when they actually differ from `editingTask.startDate/endDate`, or (b) preserve the original time-of-day when reconstructing the Date.

src/components/ops/task-list.tsx
```diff
@@ -499,6 +499,16 @@ function TaskList({ projectId, companyId, className }: TaskListProps) {
...
 
+      // Schedule fields come from the form as `YYYY-MM-DD` strings (or "").
+      // Convert to Date | null so the update mirrors the shape of
+      // ProjectTask.startDate/endDate. Sending undefined here would let the
+      // partial update skip these fields entirely — the bug (c0d4abe7) was
+      // dispatchers editing the schedule and the new dates never persisting.
+      const startDate = values.startDate
+        ? new Date(values.startDate)
+        : null;
+      const endDate = values.endDate ? new Date(values.endDate) : null;
```
jack-opsapp, line 530: Timezone footgun: `new Date("2026-05-11")` parses as UTC midnight, which renders as May 10 in any negative-UTC timezone (Pacific, Mountain, Central, Eastern, Atlantic). A dispatcher in Vancouver picks May 11 in the form, saves, and the task lands on May 10. This pattern is mirrored in `handleCreateSubmit` (line 488-489) and in the form helper `toDateInputValue` (task-form.tsx:73), so it predates this PR — but the fix propagates it to the edit path. Worth a follow-up to parse YYYY-MM-DD as local midnight (e.g. `new Date(y, m-1, d)`) on both create and edit paths so the day a user picks is the day that persists.

src/components/ops/task-list.tsx
jack-opsapp, line 541: Question — `dependencyOverrides` is also captured by TaskFormValues (task-form.tsx:34-37) but is not forwarded in `handleEditSubmit` (or `handleCreateSubmit`). If the form lets a dispatcher set per-task dependency overlap overrides, those edits are also being dropped here. If they are not editable from this surface, ignore. If they are, this PR is incomplete by the same logic that prompted the bug.

jack-opsapp:
REVIEW PASS :: CONCERN (bug 63d0cf8d) — out-of-diff finding

Stale gate not addressed by this PR. The agent-queue hooks at `src/lib/hooks/use-approval-queue.ts` still gate three queries on `usePermissionStore.can("pipeline.view")` (lines 51, 84, 98) rather than the newly-introduced `agent.queue.view`:

- `useApprovalQueue` — line 51
- `useApprovalQueueStats` — line 84
- `useApprovalQueuePendingCount` — line 98

The third one bites hardest: it is called unconditionally from `Sidebar` (sidebar.tsx:191) and has `refetchInterval: 60_000`. After this fix, every non-admin user with `pipeline.view` (e.g. role "office") will:

- NOT see the sidebar item (correct — needs `agent.queue.view`)
- NOT be able to reach the page (correct — 404 from the route gate)
- BUT still hit `/api/agent/queue?countOnly=true` every 60 seconds in the background and pile 403s into the console

The inline comment in that file specifically says it was trying to avoid this ("crew/operator users don't hit /api/agent/queue and earn 403s in the console") — but with `pipeline.view` as the gate it no longer matches the audience of the new permission.

Fix: change all three `can("pipeline.view")` to `can("agent.queue.view")` in `src/lib/hooks/use-approval-queue.ts`.

jack-opsapp:
REVIEW PASS [bug 452d7865] :: STATUS :: BLOCK MERGE

Wargamed the fix against every surface that touches `TaskStatus.InProgress`. The two-file patch (`task-form.tsx`, `task-list.tsx`) is correct and well-commented, but the fix is incomplete — a third menu surface still hands operators the same broken option, so the bug repro is still alive on the calendar.

**Blocker**
- `src/app/(dashboard)/calendar/_components/side-panel/task-detail-panel.tsx:35-40` defines its own `STATUS_OPTIONS` literal that still includes `TaskStatus.InProgress`. Open any scheduled task from the calendar grid → side panel still offers In Progress → saving silently collapses it to Booked on next refetch. Same data-loss bug, parallel surface. Apply the same omission + back-reference comment used in the other two files.

**Non-blocking (flag for follow-up bug, not this PR)**
- Five consumers branch on `status === TaskStatus.InProgress`, which is now unreachable: `getProjectComputedStatus` (models.ts:702), `crew-board-widget` (line 92), `task-pulse-widget` (128,132), `backlog-depth-widget` (116), `dashboard/page.tsx` (304). The widgets will silently under-report in-progress work; the project-status derivation will never auto-advance to InProgress. Fix is to use `deriveTaskStatusKey()` from `calendar-utils.ts` which already computes in_progress from dates.
- `FetchTasksOptions.status === TaskStatus.InProgress` filter would silently match all Booked rows via `serializeTaskStatus`. Footgun; no caller hits it today. JSDoc warning recommended.

**Verified clean**
- Round-trip Booked/Completed/Cancelled is stable.
- No existing-data migration needed — DB constraint never allowed in_progress.
- `parseTaskStatus` defensively accepts `in_progress` from DB for forward-compat.
- Null / unknown values default to Booked — reasonable.
- TanStack optimistic update + invalidation path (`useUpdateTaskStatus`) reverts cleanly.

Line-anchored details on commit 1ea85ca5.

jack-opsapp:
REVIEW PASS [bug 63d0cf8d] :: 2 concerns (1 blocker, 1 concern, 0 nitpicks)

STATUS :: BLOCK MERGE

Summary of blockers:
- The new `agent.queue.view` permission and the existing API gate `requireAdminOrOwner` use two different definitions of "admin." The permission store grants the new perm to anyone in `company.adminIds` or `company.accountHolderId`, but the API only accepts `users.role IN ('admin', 'owner')`. A user listed in `adminIds` with role=`office` would now see the sidebar entry, reach the page, and then 403 on every API call — the page would render empty/errored. Commit message claim that "non-admins now get the same 404 the API already enforces" is not true for this slice.

Secondary concern:
- `src/lib/hooks/use-approval-queue.ts` (not modified by this PR) still gates its three queries on `pipeline.view`. The 60s-interval badge query in the sidebar will now 403-spam for any pipeline.view user who is not an admin (the exact scenario the inline comment in that file said it was avoiding).

src/components/ops/task-list.tsx
```diff
@@ -300,12 +315,17 @@ function TaskRow({
...
+      {/* Overflow menu — always visible. Touch devices have no hover
+          state, so the previous `opacity-0 group-hover:opacity-100`
+          treatment made Edit/Delete unreachable on phones (bug 29fcbf09).
+          Trigger is always rendered; pointer-fine devices simply see the
+          full-opacity glyph at rest. */}
       <DropdownMenu>
         <DropdownMenuTrigger asChild>
           <button
-            className="p-[4px] text-text-mute hover:text-text-3 transition-colors shrink-0 opacity-0 group-hover:opacity-100"
+            className="p-[4px] text-text-3 hover:text-text transition-colors shrink-0"
```
jack-opsapp, line 326: The overflow trigger is now always visible (good — fix is correct), but the actual hit area is roughly 22x22px: p-[4px] around a 14px icon. The design system mandates 44x44 minimum touch targets (.interface-design/system.md line 749). On phones — the exact audience this bug exists for — field crews still won't be able to reliably tap this without zooming. Suggest min-w-[44px] min-h-[44px] (or w-11 h-11) with flex centering so the rendered glyph stays small but the tap area meets the spec. The status select trigger and avatar buttons have the same problem but are pre-existing and out of scope; the overflow trigger is touched by this commit so it's fair game.

jack-opsapp:
REVIEW PASS [bug c0d4abe7] :: STATUS :: BLOCK MERGE

The fix correctly persists startDate/endDate to the partial update, so the reported symptom (dates dropped on save) is resolved. However, the fix introduces a regression for confirmed tasks.

Findings:
1. **(BLOCKING) Spurious customer reschedule emails.** The form passes YYYY-MM-DD only. `new Date(values.startDate)` produces UTC midnight. The prior task.start_date in DB carries a real time-of-day. In task-service.ts:493-497, `startChanged` is computed by comparing ISO strings — these will never match for any task whose start time is not exactly UTC midnight. For confirmed tasks, this triggers `onConfirmedTaskRescheduled` on every edit submit, even when the dispatcher only changed status / type / color / team. Companies with `reschedule_behavior = draft` or `auto_send` will send a customer-facing reschedule email each time the form is saved. Before this PR, startDate was undefined on this path, so the hook never fired. This is a new regression. Suggest: only include startDate/endDate in the payload when they actually differ from `editingTask.startDate/endDate`, or preserve the time-of-day when reconstructing the Date.

2. **(Non-blocking, pre-existing)** Timezone parsing: `new Date('YYYY-MM-DD')` lands a day early west of UTC. Pattern is shared with handleCreateSubmit and task-form.tsx:73, so not introduced by this PR — but the fix extends the surface area.

3. **(Question)** `dependencyOverrides` is in TaskFormValues but is not forwarded by either submit handler. If it's meant to be editable from this form, it would have the same drop-the-field problem.

TanStack Query invalidation is fine — `useUpdateTask` cancels and patches tasks.detail, tasks.list, and calendar.scheduled caches with the new fields.

Date null-vs-undefined semantics are correctly handled by mapToDb (task-service.ts:214-215).

Recommend resolving (1) before merge.

src/app/(dashboard)/invoices/page.tsx
```diff
@@ -220,7 +229,34 @@ export default function InvoicesPage() {
...
+      (i) =>
+        i.projectId === projectIdFilter &&
+        i.status !== InvoiceStatus.Paid &&
+        i.status !== InvoiceStatus.Void &&
+        i.status !== InvoiceStatus.Draft,
+    );
+    if (outstanding.length === 1) {
+      setPaymentInvoice(outstanding[0]);
+      autoPaymentRanRef.current = true;
+    } else if (outstanding.length === 0 && invoices.length > 0) {
```
jack-opsapp, line 254: REVIEW PASS [bug 272b3751] :: Silent dead-end edge case. The toast fires only when `outstanding.length === 0 && invoices.length > 0`. If the company has zero invoices total (new account or first project), the user clicks Record Payment on the project canvas, gets routed here, sees a filtered-empty invoice list, and gets NO feedback explaining why nothing happened. Suggest dropping the `invoices.length > 0` guard so the toast always fires when there are no outstanding invoices for the project (covers both: no invoices at all, and no outstanding ones). Not a merge blocker, but it recreates a dead-end feel for an unlucky subset of users.

src/app/(dashboard)/invoices/page.tsx
```diff
@@ -206,8 +206,17 @@ export default function InvoicesPage() {
...
+  // Project filter pulled from the URL — set when the projects canvas
+  // hands off to the invoices page for a "Record payment" workflow
+  // (bug 272b3751). Kept reactive on every query-string change so the
+  // user can clear it via the back button without an extra re-route.
+  const projectIdFilter = searchParams.get("projectId");
+
   const filtered = useMemo(() => {
     let list = [...invoices];
+    if (projectIdFilter) {
+      list = list.filter((i) => i.projectId === projectIdFilter);
```
jack-opsapp, line 218: REVIEW PASS [bug 272b3751] :: No visible signal that the list is filtered by project. When the user arrives via `?projectId=...`, the invoice list silently hides every invoice from other projects. There is no filter chip, banner, or "Clear filter" affordance — only the back button (per inline comment above). For the multi-outstanding-invoice path (where we leave the user on the filtered list to choose), they may not realize what they are looking at. Consider rendering a dismissible chip like `// FILTERED :: {projectTitle}` with an X to clear `projectId` from the query string. Non-blocking, but improves the handoff UX.

src/app/(dashboard)/invoices/page.tsx
```diff
@@ -220,7 +229,34 @@ export default function InvoicesPage() {
...
+    );
+    if (outstanding.length === 1) {
+      setPaymentInvoice(outstanding[0]);
+      autoPaymentRanRef.current = true;
+    } else if (outstanding.length === 0 && invoices.length > 0) {
+      // No outstanding invoices to apply payment to — tell the user.
+      toast.info(t("invoices.payment.noOutstanding"));
+      autoPaymentRanRef.current = true;
+    }
+  }, [searchParams, projectIdFilter, invoices, can, t]);
```
jack-opsapp, line 259: REVIEW PASS [bug 272b3751] :: `autoPaymentRanRef` correctly prevents re-fire on subsequent renders, but it is module-scoped to this page mount, not to the URL. If a user closes the auto-opened modal without recording a payment and the URL still has `?action=recordPayment`, the modal will NOT reopen on this mount (good, prevents loop). However, if they navigate away and back via in-app routing (same React mount, same searchParams string), still no reopen. That is the intended behavior, but worth confirming: the back button from /projects to /invoices in some Next.js setups can reuse the page mount. Worth a manual test: click Record Payment, close the modal, hit browser back, hit forward — does the modal correctly NOT re-pop (desired) or does it confusingly re-pop? Either is acceptable, just choose one and verify.

src/app/(dashboard)/projects/page.tsx
```diff
@@ -174,18 +176,28 @@ export default function ProjectsPage() {
...
+  const router = useRouter();
   const { isComplete: setupComplete, needsWebSetup, missingSteps } = useSetupGate();
   const [showSetupModal, setShowSetupModal] = useState(false);
 
   // ── Permissions ──
   const canManage = can("projects.edit");
   const canViewAccounting = can("accounting.view");
   const canCreateTasks = can("tasks.create");
-  const canRecordPayment = can("accounting.edit");
+  // Record-payment lives on the invoices surface — gate the project
```
jack-opsapp, line 187: REVIEW PASS [bug 272b3751] :: Permission re-gate looks correct and matches the destination page. Previously `accounting.edit`, now `invoices.record_payment` — same permission the invoices page row-action and the auto-open effect enforce, so the action is no longer advertised to users who cannot complete it. Cross-checked: this permission is also referenced in `src/lib/feature-flags/feature-flag-definitions.ts:42`, `src/lib/types/permissions.ts:147`, and `src/lib/types/dashboard-widgets.ts:329`. Consistent. No other "record payment" entry points outside invoices and the project canvas were found.

jack-opsapp:
REVIEW PASS [bug 29fcbf09] :: Mobile task rows hide field-critical controls

**Summary of fix review (commit 701698fd):**

- Confirmed: `hidden sm:flex` removed from assignee avatars and date range — now visible at every breakpoint. Correct fix for the core bug.
- Confirmed: `opacity-0 group-hover:opacity-100` removed from overflow trigger — now always visible on touch devices. Correct fix.
- Confirmed: `aria-label="Task actions"` added — icon-only button is now accessible to screen readers. Good.
- Confirmed: color tokens used (`text-text-3`, `hover:text-text`) — no hardcoded colors introduced. Matches spec v2 text ladder.
- Confirmed: `TaskList` is only rendered from `details-tab.tsx` (project workspace). The separate `TaskRow` in `task-list-widget.tsx` is a different component and is not affected by this bug.
- Empty state confirmed safe: when no assignees the `UnassignedBadge` popover renders; when no schedule the `UnscheduledBadge` popover renders. Both reachable on mobile now.
- Long-assignee handling: `shrink-0` on the assignee block + `flex-1 truncate` on the task title means very long titles still truncate first. Acceptable.
- Permission gating preserved: Edit/Delete still wrapped in `PermissionGate`.

**Concerns (non-blocking):**

1. **Touch target size.** Overflow button is `p-[4px]` around a `14px` icon, ~22x22px. The design system requires 44x44 minimum (.interface-design/system.md line 749). See line comment on line 326. The status select trigger and assignee avatars have the same problem but are pre-existing. The overflow button is touched by this PR so it's fair to raise.
2. **Skeleton drift.** Loading skeleton at lines 373/376 (not in this diff) still uses `hidden sm:flex` / `hidden sm:block` for the avatar and date placeholders. Below 640px the skeleton looks different from the loaded row — rows visually jump on data load. Cheap to fix in the same PR by removing the `hidden sm:` prefix on both lines.

Neither concern blocks correctness of the original bug fix.

STATUS :: ALLOW MERGE

jack-opsapp:
REVIEW PASS [bug 272b3751] :: Project canvas Record Payment dead-end

**Scope verified.** This fix replaces a stub toast with a router navigation that hands off to the existing invoices page Record Payment modal. The underlying payment-writing flow (RecordPaymentModal, useRecordPayment, InvoiceService.recordPayment, payments table insert, DB-trigger-driven balance recalc) is UNCHANGED and was already working — so this is genuinely a "connect the wire" fix, not a stub-to-stub change. The end-to-end path (project canvas action -> /invoices?projectId=...&action=recordPayment -> auto-open modal -> insert into payments table -> DB trigger updates amount_paid/balance_due/status -> TanStack Query invalidates invoices.detail + invoices.lists + payments.invoicePayments) is sound.

**Wargame checklist:**
- Auth state: navigation is client-side router.push — if token expires mid-flow, useInvoices fetch fails and TanStack shows query error state; no payment write happens. Safe.
- Permissions: project canvas now gates on invoices.record_payment, matching the invoices row action and the auto-open effect. Account holders/admins get all perms via permissions-store. Consistent.
- Race conditions: useRecordPayment is a single mutation; double-click would fire twice but the modal's Submit button has no in-flight disable. NOT introduced by this PR — pre-existing behavior in RecordPaymentModal (page.tsx:854-872). Flagging for future hardening but out of scope here.
- Empty/null/extreme: payment modal already gates amount > 0 (page.tsx:869). $0/negative blocked. Currency is per-company, no mismatch concern at this layer.
- Pre-existing callers: invoices row action and project canvas are the only two entry points. Now share the same gate and the same modal. Confirmed.
- Error paths: if recordPayment mutation fails, modal stays open (onSuccess closes it) — user can retry. Toast surfacing of errors is via the mutation hook layer (not changed here).
- Server vs Client: both files are "use client". Correct.
- TanStack Query invalidation: useRecordPayment invalidates invoices.detail, invoices.lists, AND payments.invoicePayments. Correct.
- DB row writes: payments insert; DB trigger handles amount_paid/balance_due/status on invoices. RLS on payments table assumed to scope by company_id (not modified here). Out of scope for this fix.

**Concerns (all non-blocking, see line comments):**
1. Silent dead-end if company has zero invoices total (line 254 guard skips the toast).
2. No visible filter chip indicating ?projectId= is filtering the list.
3. autoPaymentRanRef behavior on browser back/forward — worth a manual smoke test.

**Bug classifier flag:** "complex by payment-classifier" — but the change itself is genuinely a routing fix on top of an already-working payment flow. The scrutiny applies, and I have applied it, and the fix is correct.

STATUS :: ALLOW MERGE

jack-opsapp:
Closing — stale (~1 month) and now conflicting with main; the bulk is superseded by intervening work. The still-live fixes this sweep surfaced (critical Stripe default-payment-method; calendar multi-day overlap) are being re-implemented fresh on current main. Reference commits remain available via this PR's refs/pull/52/head.

PR #53, opened 2026-05-11
Nightly Bug Sweep — web — 2026-05-11 — Batch 02

jack-opsapp:
Nightly QA sweep for OPS Web. Ten bugs in this batch covering notification reliability, calendar correctness, billing recovery, and inbox/email wizard UX. Each commit is scoped to one bug.

## Fixes

- **119ddcde** · functional · low — Project mention notifications were fire-and-forget with only `console.error` on failure. Dispatch helpers now return `DispatchResult`; `useCreateProjectNote` awaits in-app + push channels and surfaces failures via a toast warning on the activity tab. Note save is unaffected.
- **3f4e77eb** · functional · high · **COMPLEX** — `fetchScheduledTasksForRange` filtered `start_date BETWEEN`, so multi-day tasks beginning before the visible window disappeared from the calendar. Added `overlapRangeStart`/`overlapRangeEnd` options applying a Postgres range-overlap predicate. Single-day tasks still surface via OR clause.
- **3258513c** · functional · high · **COMPLEX** — Email import wizard kept discarded leads in the payload (imported them with stage="discarded") despite the confirm step copy promising they wouldn't be imported. Effective stage now computed in the importLeads filter and discarded leads dropped before serialization.
- **e37ee17f** · ui · medium — Email analysis error UI was a wall with no exit. Extracted `startAnalysis` into a `useCallback`; added `handleRetry` that resets poll/interpolate timers and per-run state; rendered Retry + optional Back-to-connection actions on the error panel.
- **c7c8414d** · ui · medium — Inbox `ThreadDetailView` destructured `data/isLoading` only, so query errors looked like empty threads. Now reads `error/refetch/isRefetching` and renders a failed-detail alert with a spinning Retry button inside the message scroll area.
- **31b9ed35** · ui · medium · **COMPLEX** — Calibration page returned null on permission failure or missing deck, producing a blank screen. Added page-level permission check (email.configure_ai) with ACCESS DENIED panel, query-error panel with retry, and DECK NOT FOUND panel with reload.
- **d161843d** · functional · medium — Persistent "Pipeline import complete" notification stayed pinned to the rail after activation. Activate route now resolves matching notifications (is_read=true) by user+company+title once the connection flips active; wrapped in its own try/catch so cleanup hiccups never fail activation.
- **990491b0** · functional · high · **COMPLEX** — Re-scan button forced active connections to the activation confirmation step instead of refreshing leads. Added opt-in `rescan` prop on `ImportPipelineWizard`; when true + active, the wizard resets state and lands on step 2 so `AnalyzeStep` auto-fires a new /analyze call. integrations-tab passes `rescan: true` from the re-scan button only.
- **a0edae0f** · functional · critical · **COMPLEX** — SetupIntent attached a card but never set it as default, so subscribe/recover flows kept failing for locked customers. New POST /api/stripe/payment-methods action='set_default' attaches (idempotent) and updates `invoice_settings.default_payment_method`. AddCardForm auto-promotes when no existing default; non-default cards expose explicit "Set as default" action.
- **554c4cf8** · ui · medium · **COMPLEX** — Billing screens drifted from OPS spec. Removed exclamation from `subscription.toast.subscribed` (EN+ES); trimmed sentimental "Welcome aboard" on Checkout return; replaced `rounded-full` pill on current-plan and invoice-status badges with 4px chip radius per spec; moved hardcoded "Cancel" into the settings dictionary (EN+ES).

## Test plan

- [ ] Mention a teammate while notification API returns 500 → toast warning fires, note still saves
- [ ] Open calendar on a week mid-way through a multi-day task → task appears
- [ ] Run email import with at least one discarded lead → discarded leads do not appear in opportunities
- [ ] Force /analyze to 500 → Retry button on the error panel reruns the analysis without remounting
- [ ] Force inbox thread fetch to error → failed-detail alert with Retry surfaces
- [ ] Open /calibration without email.configure_ai → ACCESS DENIED page; with permission but deck 500 → DECK UNAVAILABLE with Retry
- [ ] Complete email import + click Activate → "Pipeline import complete" notification disappears from the rail
- [ ] On an active connection, click re-scan → wizard lands on AnalyzeStep with a fresh job (not step 5)
- [ ] Lockout state + add card via AddCardForm → card promoted to default, subscribe succeeds
- [ ] Visit /settings billing tab → no rounded-full pills, "Plan updated" toast (no exclamation), Cancel label localized


src/lib/api/services/task-service.ts
```diff
@@ -311,6 +319,24 @@ export const TaskService = {
...
+      // Single-day tasks (end_date IS NULL) still surface because the
+      // start_date upper bound is the rangeEnd; the OR catches them when
+      // their start sits inside the window.
+      const rangeStartISO = options.overlapRangeStart.toISOString();
+      const rangeEndISO = options.overlapRangeEnd.toISOString();
+      query = query
+        .lte("start_date", rangeEndISO)
+        .or(
+          `end_date.gte.${rangeStartISO},and(end_date.is.null,start_date.gte.${rangeStartISO})`,
+        );
```
jack-opsapp, line 337: Range-overlap predicate looks correct: covers start-before/end-inside, start-inside/end-after, fully-spanning, and single-day-inside cases. Null end_date is preserved via the OR clause. Cache key in use-tasks.ts still uniquely identifies the range. Good.

src/lib/api/services/task-service.ts
```diff
@@ -311,6 +319,24 @@ export const TaskService = {
       query = query.lte("start_date", options.startDateTo.toISOString());
     }
 
+    if (options.overlapRangeStart && options.overlapRangeEnd) {
```
jack-opsapp, line 322: Nit (non-blocking): the doc above says overlapRange* and startDateFrom/startDateTo are 'mutually exclusive — use one or the other,' but fetchTasks doesn't enforce it. If a future caller passes both, you'll get two competing 'lte("start_date")' filters and a confusing result. fetchScheduledTasksForRange's Omit prevents this today, but a defensive guard (or dev-time warn) inside fetchTasks would future-proof the contract.

jack-opsapp:
REVIEW PASS [bug 3f4e77eb] :: Range-overlap predicate is correct across all four overlap cases (start-before/end-inside, start-inside/end-after, fully-spanning, single-day-inside). Null end_date is preserved via the OR. Pagination (cursor + remaining) still works because count is computed against the same filtered query. Cache key in useScheduledTasks still uniquely identifies the range (startStr+endStr). Only caller of fetchScheduledTasksForRange is useScheduledTasks — no callers depend on the old narrow start_date BETWEEN semantics. One non-blocking nit posted inline about the mutual-exclusion contract between overlapRange* and startDateFrom/To not being enforced inside fetchTasks. STATUS :: ALLOW MERGE.

src/app/(dashboard)/calibration/page.tsx
```diff
@@ -34,6 +79,43 @@ export default function CalibrationPage() {
...
+    return (
+      <div className="px-11 py-9 max-w-[840px] mx-auto">
+        <div className="flex items-center gap-2 mb-4">
+          <AlertCircle className="w-[18px] h-[18px] text-[#B58289]" />
+          <h1 className="font-cakemono font-light uppercase text-[22px] text-text">
+            DECK UNAVAILABLE
+          </h1>
+        </div>
+        <p className="font-mohave text-[14px] text-text-2 mb-2">
+          The calibration deck failed to load. The service may be
```
jack-opsapp, line 99: REVIEW PASS [bug 31b9ed35] :: This panel renders for any deck/firstRun failure — 403 (permission), 401 (session expired), 500 (server), and network errors all collapse into the same DECK UNAVAILABLE copy. The hook in `use-calibration-deck.ts:20` throws a generic Error with no status code, so distinguishing 403/401/500 here would require widening the hook to attach `res.status` to the thrown error. Today the page-level `hasConfigureAi` gate above shields the common 403 path, so this is acceptable as-is — but if the API ever 403s for a user who passes the local permission check (stale cache, role just revoked), they'll see DECK UNAVAILABLE instead of ACCESS DENIED. Consider a follow-up to surface status.

src/app/(dashboard)/calibration/page.tsx
```diff
@@ -21,8 +22,52 @@ export default function CalibrationPage() {
...
 
-  const { data: deck, isLoading: deckLoading } = useCalibrationDeck();
-  const { data: firstRun, isLoading: firLoading } = useCalibrationFirstRun();
+  const can = usePermissionStore((s) => s.can);
+  const hasConfigureAi = can("email.configure_ai");
+
+  const {
+    data: deck,
+    isLoading: deckLoading,
+    error: deckError,
```
jack-opsapp, line 31: REVIEW PASS [bug 31b9ed35] :: Heads up — `useCalibrationDeck` does not override `retry`, so TanStack Query's default `retry: 3` applies. A 403 from `/api/calibration/deck` will retry 3x (with exponential backoff) before the error finally surfaces here, and `refetchInterval: 30_000` keeps polling forever even while erroring. Permission failures should not retry. Recommend adding `retry: (failureCount, err) => failureCount < 2 && !(err instanceof Error && /403|401/.test(err.message))` to the hook, or better, throwing a typed error with `.status`. Not blocking — the page-level permission gate already short-circuits before the fetch fires.

src/app/(dashboard)/calibration/page.tsx
```diff
@@ -55,7 +137,34 @@ export default function CalibrationPage() {
...
+          </h1>
+        </div>
+        <p className="font-mohave text-[14px] text-text-2 mb-6">
+          We could not assemble a calibration deck for this company. This
+          is usually fixed by reloading; if it persists, contact support.
+        </p>
+        <button
+          type="button"
+          onClick={() => refetchDeck()}
+          disabled={deckRefetching}
```
jack-opsapp, line 158: REVIEW PASS [bug 31b9ed35] :: Good — distinguishing DECK NOT FOUND (warning amber #C4A868) from DECK UNAVAILABLE (rose #B58289) and ACCESS DENIED is exactly the three-way split the bug called for. Three things worth noting: (1) the 'deck loaded but has 0 cards' case is NOT handled here — `!deck` is false when deck is `{ cards: [] }`, so it falls through to `<CommandDeck>` rendering. Out of scope for this bug but worth flagging. (2) Auth state mid-load: if the user signs out while loading, `enabled: !!companyId` becomes false but the cached `deck` is retained, so no crash — graceful. (3) Tone of copy is consistent with spec v2 (Cake Mono Light uppercase titles, font-mohave body, tactical brackets for metadata). LGTM.

jack-opsapp:
REVIEW PASS [bug 31b9ed35] :: STATUS :: ALLOW MERGE. The fix replaces the blank-screen failure modes with three distinct recoverable states: ACCESS DENIED (rose, permission), DECK UNAVAILABLE (rose, query error + retry), DECK NOT FOUND (amber, missing deck + reload). Order of checks is correct — permission gate runs before any data fetch, so users without `email.configure_ai` never trigger the API. Copy and tokens follow spec v2 (Cake Mono Light uppercase titles, font-mohave body, accent-bordered outlined buttons, tactical brackets for metadata). Non-blocking follow-ups: (1) the underlying `useCalibrationDeck` hook throws a generic Error with no status code, so 403/401/500 collapse into the same DECK UNAVAILABLE panel — consider attaching `res.status` for finer branching. (2) TanStack Query's default `retry: 3` still applies — permission errors should not retry; recommend overriding `retry` in the hook. (3) The 'deck loaded but has 0 cards' case still falls through to `<CommandDeck>` — out of scope for this bug. Auth sign-out mid-load degrades gracefully via `enabled: !!companyId` + cache retention.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -301,14 +310,42 @@ export function ImportPipelineWizard({
...
+          setConfirmedSources([]);
+          setConfirmedLeads([]);
+          setImportResult(null);
+          setImportJobId(null);
+          setRunningJobId(null);
+          setRunningJobType(null);
+          setBgProgress({ percent: 0, message: "Working..." });
+          setDirection(1);
+          setStep(2);
+          return;
```
jack-opsapp, line 331: REVIEW PASS [bug 990491b] :: rescan branch correctly null-routes existingJobId here, so AnalyzeStep's mount effect (startedRef gate) fires a fresh POST /api/integrations/email/analyze instead of resuming a stale jobId. Side-state resets (analysisResult, confirmed*, importResult, importJobId, runningJobId, runningJobType, bgProgress) are comprehensive — no residual step-5 data can leak back into the new flow. setStateCheckComplete(true) still fires in the finally block at line 599, so the step-2 spinner clears.

src/components/settings/subscription-tab.tsx
```diff
@@ -344,7 +347,7 @@ export function SubscriptionTab() {
 
     if (result === "success") {
       toast.success("Subscription active", {
-        description: "Welcome aboard. Refreshing your account…",
+        description: "Refreshing account state.",
```
jack-opsapp, line 350: REVIEW PASS [bug 554c4cf8] :: Hardcoded English strings remain in this same Checkout-return effect: "Subscription active" and "Refreshing account state." (and "Checkout cancelled" at L355). The commit explicitly justifies de-exclamating subscription.toast.subscribed by routing through the dictionary — but this companion toast (fired immediately after a successful Stripe return, i.e., the same "highest-trust moment" the bug calls out) bypasses i18n entirely. Spanish users will see English. Recommend hoisting these three strings to subscription.toast.checkoutActive / .checkoutActiveDesc / .checkoutCancelled in en + es dictionaries. Non-blocking but completes the audit.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -301,14 +310,42 @@ export function ImportPipelineWizard({
...
         // connections, regardless of what the job IDs look like.
-        if (filters.wizardCompleted && conn.status === "active") {
+        //
+        // EXCEPTION — re-scan: the integrations tab's "re-scan" button
+        // explicitly asks to refresh leads on an active connection. In
+        // that mode we skip this jump-to-step-5 path and start a fresh
+        // analyze run at step 2 (the AnalyzeStep auto-starts a new job
+        // because lastScanJobId stays untouched but startedRef on the
+        // step is fresh on mount).
+        if (!rescan && filters.wizardCompleted && conn.status === "active") {
```
jack-opsapp, line 348: REVIEW PASS [bug 990491b] :: !rescan gate on the active+wizardCompleted early return is the right place to slot the bypass — the loadedImportResult fallback path below stays intact for normal reopens (initial setup, sync-history click). Only the explicit re-scan button passes rescan:true; all six other openWizard() callers in integrations-tab.tsx (lines 434, 451, 493, 503, 541, 569) default to the legacy reconnect-to-step-5 behavior. State machine is idempotent: closing the wizard resets wizardRescan to false via onOpenChange, so a second open without re-clicking re-scan reverts to default.

src/components/settings/integrations-tab.tsx
```diff
@@ -334,11 +342,16 @@ export function IntegrationsTab() {
...
+        onOpenChange={(o) => {
+          setWizardOpen(o);
+          if (!o) setWizardRescan(false);
+        }}
         connectionId={companyConnections[0]?.id}
         companyId={companyId}
+        rescan={wizardRescan}
         onComplete={async () => {
           setWizardOpen(false);
+          setWizardRescan(false);
```
jack-opsapp, line 354: REVIEW PASS [bug 990491b] :: wizardRescan is reset in three places: onOpenChange close (line 347), onComplete (line 354), and openWizard itself only sets true when opts.rescan is true. Closing-and-reopening from another entry point (sync-history click, prompt) cannot accidentally inherit rescan=true. Post-rescan invalidation/refetch of queryKeys.gmailConnections.all in onComplete is awaited sequentially so the active tile rerenders with updated syncFilters.lastScanJobId before next render. No additional invalidation needed for inbox/threads — the analyze pipeline writes to email_connections, which is the only client-side query consumed by this surface. OAuth token refresh is handled server-side in /api/integrations/email/analyze (not in this client diff).

src/components/settings/subscription-tab.tsx
```diff
@@ -270,7 +270,10 @@ function PlanCard({
                 {info.displayName}
               </h4>
               {isCurrent && (
-                <span className="font-mono text-micro text-text bg-[rgba(255,255,255,0.08)] px-[6px] py-[2px] rounded-full uppercase tracking-wider shrink-0">
+                <span
+                  className="font-mono text-micro text-text bg-[rgba(255,255,255,0.08)] px-[6px] py-[2px] uppercase tracking-wider shrink-0"
+                  style={{ borderRadius: 4 }}
```
jack-opsapp, line 275: REVIEW PASS [bug 554c4cf8] :: Inline `style={{ borderRadius: 4 }}` works but is the only place in the file that escapes hatch out to inline styles instead of Tailwind utilities — the design system has `rounded-sm` (4px) which the lockout-overlay already uses for badges (see `lockout-overlay.tsx:333`). Swapping to `rounded-sm` keeps token alignment + dedups how chip radii are expressed across billing surfaces. Same applies to the matching change in `billing-tab.tsx:256`. Non-blocking.

src/app/api/stripe/payment-methods/route.ts
```diff
@@ -69,6 +76,89 @@ export async function GET(req: NextRequest): Promise<NextResponse> {
   }
 }
 
+export async function POST(req: NextRequest): Promise<NextResponse> {
```
jack-opsapp, line 79: REVIEW PASS [bug a0edae0 / CRITICAL] :: **No authentication on POST.** `setup-intent/route.ts` and `subscribe/route.ts` both decode the Firebase bearer token via `decodeFirebaseToken(authHeader)` before touching Stripe. This new POST handler accepts `{ companyId, paymentMethodId, action }` from any unauthenticated caller. Since `paymentMethods.attach` rejects cross-customer PMs, the worst-case is limited to: an attacker who knows a victim companyId AND a PM id already on that customer can flip the victim`s default PM to a non-default card on their own account (e.g. an expired card). That degrades into a billing failure for the victim. Add the same Firebase token decode + verify the token`s user actually owns/admins `companyId` (look at how `subscribe` correlates `user.uid` to the company — it currently does NOT, which is its own latent issue). Minimum fix: require auth header; reject if missing. Better: confirm caller is a billing-permitted member of `companyId`.

src/components/settings/billing-tab.tsx
```diff
@@ -191,7 +253,10 @@ function InvoiceStatusBadge({ status }: { status: string | null }) {
   };
 
   return (
-    <span className={`font-mono text-micro uppercase tracking-wider px-[6px] py-[2px] rounded-full ${className}`}>
+    <span
+      className={`font-mono text-micro uppercase tracking-wider px-[6px] py-[2px] ${className}`}
+      style={{ borderRadius: 4 }}
+    >
```
jack-opsapp, line 259: REVIEW PASS [bug 554c4cf8] :: Same `team-tab.tsx:645,719` ships chip badges with the old rounded-full pill treatment (font-mono micro uppercase, px-[6px] py-[2px]) — the exact pattern this PR just corrected in billing/subscription. If the audit principle is "spec 4px chip radius," team-tab membership/status chips are the next obvious carrier of the same drift. Not in scope for this commit but flagging so it lands in a follow-up rather than getting forgotten.

jack-opsapp:
REVIEW PASS [bug 990491b] :: Fix verified. Re-scan correctly bypasses the active-connection guard via opt-in rescan prop. Flow: integrations-tab passes rescan:true only from the re-scan button (line 597); wizard's checkWizardState resets all derived state (existingJobId, analysisResult, confirmedSources/Leads, importResult, importJobId, runningJobId, runningJobType, bgProgress) and lands on step 2; AnalyzeStep's startedRef-gated mount effect auto-fires POST /api/integrations/email/analyze because existingJobId is null. State machine is idempotent — wizardRescan resets on close (onOpenChange) and on onComplete. Other openWizard() callers (6 sites) keep legacy reconnect-to-step-5 behavior. Error path: AnalyzeStep renders inline error UI with retry on analyze failure (line 256-266). Cache invalidation: queryKeys.gmailConnections.all is invalidated AND refetched sequentially in onComplete — sufficient since this surface only consumes email_connections data. OAuth token refresh stays server-side in /api/integrations/email/analyze (unchanged by this diff). No concerns.\n\nSTATUS :: ALLOW MERGE

src/app/api/stripe/payment-methods/route.ts
```diff
@@ -69,6 +76,89 @@ export async function GET(req: NextRequest): Promise<NextResponse> {
...
+    try {
+      await stripe.paymentMethods.attach(paymentMethodId, {
+        customer: customerId,
+      });
+    } catch (attachErr) {
+      const msg =
+        attachErr instanceof Error ? attachErr.message : String(attachErr);
+      // Stripe code 'resource_already_exists' means it's already attached
+      // to this customer — that's fine.
+      if (!msg.toLowerCase().includes("already")) {
```
jack-opsapp, line 134: REVIEW PASS [bug a0edae0 / CRITICAL] :: **Brittle error-string matching.** Swallowing the attach error by checking `msg.toLowerCase().includes("already")` is fragile. Stripe surfaces this case as a `StripeInvalidRequestError` with `code: "resource_already_exists"` (or in some PM-attach cases the API actually returns 200 idempotently — needs verification). Match on `err instanceof Stripe.errors.StripeInvalidRequestError && err.code === "resource_already_exists"`. If Stripe ever localizes or changes the message wording, your happy path silently flips into a 400 returned to the client right after the user clicks `Set as default` — i.e., the exact lockout symptom this PR is supposed to fix. Worth tightening before merge.

src/components/settings/billing-tab.tsx
```diff
@@ -115,16 +148,45 @@ function AddCardForm({ onSuccess, onCancel }: { onSuccess: () => void; onCancel:
...
+      // customer default — adding a card via SetupIntent alone leaves
+      // the customer locked out.
+      const newPaymentMethodId =
+        typeof setupIntent?.payment_method === "string"
+          ? setupIntent.payment_method
+          : setupIntent?.payment_method?.id ?? null;
+      const hasExistingDefault = (existingMethods ?? []).some((m) => m.isDefault);
+
+      if (newPaymentMethodId && !hasExistingDefault) {
+        try {
```
jack-opsapp, line 172: REVIEW PASS [bug a0edae0 / CRITICAL] :: **Cache-race auto-promotion.** `usePaymentMethods()` may still be loading the first time AddCardForm renders — `existingMethods` is `undefined`, so `(existingMethods ?? []).some(...)` is `false`, and the new card is silently promoted to default. That auto-promotes when there IS already a default but the cache just hadn`t hydrated. In practice this is rarely user-visible (the rail/subscribe flow only cares that *some* default exists), and arguably aligns with intent ("the card the operator just added is the one they want billed"). Worth either: (a) awaiting a fresh `refetch()` before deciding, or (b) documenting that the new card always becomes default when the prior default cache is unknown. As-is, leave a code comment so the next reader doesn`t treat this as accidental.

jack-opsapp:
REVIEW PASS [bug 554c4cf8] :: STATUS :: ALLOW MERGE

Surfaced drifts in the audit (exclamation in subscription.toast.subscribed EN+ES, rounded-full pill on current-plan + invoice-status badges, "Welcome aboard" sentimentality on Checkout return, hardcoded "Cancel" label) are all correctly resolved. Voice now reads as command-deck calm. Amber preserved only on legitimate semantic warnings (trial countdown line 437, open-invoice status line 239) which maps correctly to the tan/amber semantic token — verified.

Three non-blocking follow-up notes posted as line comments:
1. Same Checkout-return effect still has THREE hardcoded English strings ("Subscription active", "Refreshing account state.", "Checkout cancelled") that bypass i18n entirely — Spanish users will see English at the same high-trust moment the bug calls out.
2. The 4px chip radius is applied via inline `style={{ borderRadius: 4 }}` rather than the Tailwind `rounded-sm` utility the rest of the system uses (e.g. lockout-overlay.tsx:333).
3. team-tab.tsx:645,719 still ship the rounded-full pill pattern this PR just corrected — same drift, different surface.

None of these are introduced by this commit; they are pre-existing audit residue. ALLOW MERGE; recommend a Pass-2 ticket to close them out.

src/components/settings/billing-tab.tsx
```diff
@@ -69,10 +82,28 @@ function PaymentMethodCard({ method, onRemove, isRemoving }: { method: PaymentMe
...
+            onClick={() => onSetDefault(method.id)}
+            disabled={isSettingDefault}
+            className="font-mono text-micro uppercase tracking-wider px-[6px] py-[2px] text-text-2 hover:text-ops-accent border border-white/10 hover:border-ops-accent transition-colors disabled:opacity-50 disabled:cursor-not-allowed"
+            style={{ borderRadius: 4 }}
+            title="Set this card as the default for subscription charges"
+          >
+            {isSettingDefault ? (
+              <Loader2 className="w-[12px] h-[12px] animate-spin inline" />
+            ) : (
+              "Set as default"
```
jack-opsapp, line 104: REVIEW PASS [bug a0edae0 / CRITICAL] :: **i18n regression.** Hardcoded English strings introduced: `"Set as default"` button label, `title="Set this card as the default for subscription charges"`, and in `BillingTab`: `"Default card updated"`, `"Future subscription charges will use this card."`, `"Could not set default card"`, `"Unknown error"`. Also the warning toast description in AddCardForm: `"Card added but could not be set as default automatically..."`. Project rule (CLAUDE.md): all user-facing strings via `useDictionary("settings")`. Add keys like `billing.setAsDefault`, `billing.setAsDefault.tooltip`, `billing.toast.defaultUpdated{.title,.description}`, `billing.toast.defaultFailed{.title,.description}` to both `en/settings.json` and `es/settings.json` before merge.

src/app/api/stripe/payment-methods/route.ts
```diff
@@ -69,6 +76,89 @@ export async function GET(req: NextRequest): Promise<NextResponse> {
...
+        // If it's attached to a different customer, surface the error.
+        console.error("[stripe/payment-methods] attach error:", attachErr);
+        return NextResponse.json({ error: msg }, { status: 400 });
+      }
+    }
+
+    // Promote to default for invoices + future subscription charges.
+    await stripe.customers.update(customerId, {
+      invoice_settings: { default_payment_method: paymentMethodId },
+    });
```
jack-opsapp, line 144: REVIEW PASS [bug a0edae0 / CRITICAL] :: **Customer default only — open subscription not retried.** The fix sets `invoice_settings.default_payment_method` on the *customer*. That correctly unblocks the documented flow: lockout → add card → call `/subscribe` again → subscribe sees the customer default and skips the 402 (`subscribe/route.ts:218-228`). But: if the company already has a `past_due` / `unpaid` subscription whose latest invoice is still open, this fix does NOT (a) update `subscription.default_payment_method` for the existing sub, or (b) retry/pay the open invoice. Those customers will not automatically exit `past_due` — they need either Stripe`s smart retries to re-attempt with the new customer default, or a manual `invoice.pay()` call. Confirm with the team whether "recovery" in this bug means "let the operator re-subscribe from scratch" (this PR solves it) or "resume the existing past_due sub" (this PR does NOT solve it). If the latter, add: `await stripe.subscriptions.update(subId, { default_payment_method: paymentMethodId })` and `await stripe.invoices.pay(latestOpenInvoiceId)` here.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -729,11 +768,28 @@ export function ImportPipelineWizard({
...
+      const decision = triageDecisions.get(lead.id);
+      const effectiveStage =
+        decision === "won"
+          ? "won"
+          : decision === "lost"
+            ? "lost"
+            : decision === "discard"
+              ? "discarded"
+              : lead.stage;
+      if (effectiveStage === "discarded") return false;
```
jack-opsapp, line 792: Parity gap with the confirm step. confirm-pipeline-step.tsx getEffectiveStage() treats decision=='active' as 'lead.stage is terminal? new_lead : lead.stage'. The filter here doesn't replicate that override: a lead the user explicitly marked 'active' but whose AI stage is 'discarded' will fall into the lead.stage branch, get effectiveStage='discarded', and be silently dropped — even though the UI groups it under New Lead and counts it in 'active'. Mirror the confirm-step logic exactly or extract a shared resolveEffectiveStage(lead, triageDecisions) helper used by both.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -729,11 +768,28 @@ export function ImportPipelineWizard({
...
       return true;
     });
```
jack-opsapp, line 794: Empty payload edge case: if the operator discards every lead, importLeads.length === 0, the wizard POSTs {leads: []} and the server route returns 400 'connectionId, companyId, and leads required'. The wizard catches that and surfaces a generic toast.error('Import failed (400)') — confusing UX for what is actually a legitimate user choice. Guard before submit: if importLeads.length === 0, show a deliberate empty-state message ('Nothing to import — every lead was discarded') and either short-circuit to step 5 or bounce back to triage. Don't let a legitimate user choice surface as an error.

jack-opsapp:
REVIEW PASS [bug a0edae0 / CRITICAL] :: Reviewed commit `ecb98d9` end-to-end against `/api/stripe/setup-intent`, `/api/stripe/subscribe`, `useSetDefaultPaymentMethod`, and the existing `payment-methods` GET/DELETE.

**Core fix is correct for the documented flow.** SetupIntent attached but never promoted → `subscribe` 402 → lockout. The new `POST { action: 'set_default' }` writes `customer.invoice_settings.default_payment_method`, which is exactly what `subscribe/route.ts:218` checks. The new `AddCardForm` auto-promote and per-card "Set as default" both round-trip through the new endpoint. Idempotent re-attach is intentional. UI is permission-gated client-side via `can("settings.billing")`. Stripe call shape is correct.

**Concerns (line comments above):**
1. **POST has no auth header check** — sibling routes (`setup-intent`, `subscribe`) decode the Firebase bearer token; this one accepts any caller. Cross-customer PM attach is rejected by Stripe so blast radius is bounded, but a caller who knows a victim companyId + an already-attached PM id on that customer can flip their default.
2. **Brittle `msg.includes("already")`** error swallow — Stripe could change the wording or localize and silently break the happy path.
3. **Cache-race auto-promotion** in `AddCardForm` — when `usePaymentMethods()` is still loading, `hasExistingDefault` is false and we auto-promote regardless. Probably fine; document it.
4. **i18n regression** — `"Set as default"`, the title attr, and four new toast strings in `BillingTab` are hardcoded English. Project rule says all user-facing strings go through `useDictionary`.
5. **Recovery scope** — fixes "re-subscribe after lockout" but does NOT update `subscription.default_payment_method` on an existing `past_due` sub, nor retry the open invoice. If "recovery" means resuming an existing past_due subscription rather than starting a new one, this PR is incomplete.

None of these are showstoppers for the symptom in the bug. (1) is the only one I'd insist on before merge for a CRITICAL billing fix.

STATUS :: BLOCK (auth-on-POST + i18n must land before merge; the other three are strong recommendations)

src/components/settings/import-pipeline-wizard.tsx
jack-opsapp, line 794: Two follow-ups outside this diff's window but on the same code path:\n\n(1) Payload stage resolver at lines 805-808 duplicates this filter's stage-resolution ternary. Today they happen to stay in sync because the filter strips discards before the payload builder runs, but two copies of the same logic invite future drift. Extract once (resolveEffectiveStage(lead, triageDecisions)) and call twice.\n\n(2) TanStack invalidation gap: after the import job kicks off (line 864) and again when handleImportComplete fires (line 878) only invalidateConnections() runs, which only touches queryKeys.gmailConnections. The opportunities / pipeline list queries are NOT invalidated. After the user finishes the wizard and navigates to /pipeline, they see stale data until a manual refetch. Add queryClient.invalidateQueries({ queryKey: queryKeys.opportunities.all }) (and clients.all) in handleImportComplete.

src/components/settings/import-pipeline-wizard.tsx
jack-opsapp, line 792: Defense-in-depth observation (server side, not in this PR's diff): /api/integrations/email/import/route.ts blindly trusts whatever 'stage' the client sends — its stageMap (line 322-334) accepts 'discarded' and creates an opportunity with OpportunityStage.Discarded. Today the wizard is the only caller, so the client-side filter is sufficient. But if a future caller (or a manually-crafted request) sends stage='discarded', it'll be silently accepted. Consider rejecting stage==='discarded' at the route boundary as a server-side guard, so the UI promise holds regardless of how the route is called.

jack-opsapp:
REVIEW PASS [bug 3258513c] :: Core fix lands the contract: discarded leads no longer reach the import payload, so the confirm-step copy ('won't be imported') now matches actual behavior. Single entry point confirmed — only the wizard hits /api/integrations/email/import. Concerns are non-blocking polish:

1. **Parity drift risk** — the wizard's effectiveStage resolver does NOT replicate confirm-pipeline-step's 'active triage on AI-discarded → new_lead' override. Result: an explicitly-triaged 'active' lead with AI stage='discarded' gets silently dropped instead of imported as new_lead. Extract a shared resolver.
2. **Empty payload UX** — discarding every lead → POSTs {leads:[]} → server 400 → generic toast. Add a pre-submit guard for a clean empty state.
3. **TanStack stale-data gap** — handleImportComplete invalidates gmailConnections only. Opportunities/clients lists stay stale after import.
4. **Defense in depth** — server route blindly accepts stage='discarded'. Worth adding a route-level reject as a server-side guard.

None of these reintroduce the original bug. Fix is correct in scope.

STATUS :: ALLOW MERGE

jack-opsapp:
Closing — stale (~1 month) and now conflicting with main; the bulk is superseded by intervening work. The still-live fixes this sweep surfaced (critical Stripe default-payment-method; calendar multi-day overlap) are being re-implemented fresh on current main. Reference commits remain available via this PR's refs/pull/53/head.

PR #54, opened 2026-05-11
Nightly Bug Sweep — web — 2026-05-11 — Batch 03

jack-opsapp:
## Summary

Batch 03 of 04 for the 2026-05-11 nightly web bug sweep. Nine of ten bugs fixed; one escalated (depends on missing module already escalated in batch-01).

### 59bfdae9 — Standalone /locked page always shows expired-subscription pricing  COMPLEX
- category: functional · severity: medium · file: src/app/(auth)/locked/page.tsx
- Standalone /locked rendered the admin pricing grid unconditionally; never called getLockoutReason, never inspected role.
- Restructured LockedPage to mirror LockoutOverlay's four-state matrix (admin-expired pricing / member-expired request / admin-unseated team CTA / member-unseated request) and fall back to pricing while auth/company still loading.

### 3f27ae82 — Billing tab hides load failures as empty states
- category: ui · severity: medium · file: src/components/settings/billing-tab.tsx
- usePaymentMethods / useStripeInvoices errors fell through to "No payment method" / "No billing history".
- Surface isError/error/isFetching/refetch from both queries; render an error block with retry button between loading and empty branches. Adds en/es billing.error.* keys.

### e67562e5 — Cancel at period end locks customers out immediately  COMPLEX
- category: functional · severity: high · file: src/app/api/stripe/cancel/route.ts
- Route set cancel_at_period_end=true on Stripe AND immediately wrote subscription_status='cancelled' — locking the customer out instantly even though they paid for the rest of the period.
- Removed the immediate DB write; the Stripe webhook (customer.subscription.deleted) remains the authoritative writer when the period actually ends. Route returns cancelAt ISO so the UI can show "Cancels on <date>".

### 9f55103f — Trial upgrade blocked by subscribe guard  COMPLEX
- category: functional · severity: critical · file: src/app/api/stripe/subscribe/route.ts
- Double-subscribe guard rejected status in ('active','trial','grace'). 'trial' is the free in-app trial — no Stripe subscription exists yet, so trial users could never convert to paid.
- Dropped 'trial' from rejected statuses; 'active' and 'grace' still rejected (real paid subs). Stripe idempotency-key already protects against in-flight double-submits.

### f920326c — Billing recovery shows success before subscription is current  COMPLEX
- category: functional · severity: high · file: src/components/settings/subscription-tab.tsx
- ?result=success fired "Subscription active" + refetched once. Webhook lag left the lockout overlay up while the toast claimed success.
- Polls company state every 2s up to 30s, pushes each fetched company into the auth store so LockoutOverlay re-evaluates, only swaps to the success toast when status is active/grace AND plan is a paid tier. Soft "still finalizing" message on timeout.

### 728027bb — Critical workflows bypass token refresh helper  COMPLEX
- category: functional · severity: high · file: src/lib/utils/authed-fetch.ts
- Plain fetch in billing fetches, manual email sync, and the import-pipeline wizard's save/import/activate handlers silently 401'd when the Firebase token aged out.
- Converted: use-billing.ts (usePaymentMethods, useStripeInvoices, useRemovePaymentMethod), use-email-connections.ts (useTriggerEmailSync), import-pipeline-wizard.tsx (saveReviewState, checkWizardState, handleImport, handleActivate). authedFetch already refreshes-and-retries on 401.

### a0bd0021 — Estimate/invoice Send does not actually email customers  COMPLEX
- category: functional · severity: critical · files: src/components/ops/send-estimate-flow.tsx, services, hooks, new API routes
- sendEstimate / sendInvoice services flipped status='sent' and wrote sent_at without ever calling SendGrid. Customers never received the document while OPS marked it sent.
- New POST /api/estimates/[id]/send and /api/invoices/[id]/send routes (verifyAdminAuth) load doc + client + company + portal branding, mint a PortalAuthService token, send via existing sendEstimateReady/sendInvoiceReady, then write status='sent'. Service methods delegate to the routes; hook mutation inputs widened to id|{id,email}; SendEstimateFlow passes the typed email. No new env vars introduced.

### ce6495b9 — Estimates/invoices can be saved empty or invalid  COMPLEX
- category: data · severity: critical · files: src/app/(dashboard)/estimates/page.tsx, src/app/(dashboard)/invoices/page.tsx
- Forms mapped line items straight to a Partial<Create*> payload and submitted; empty client / blank line items / zero totals all saved.
- Wired the existing createEstimateSchema / createInvoiceSchema (which require companyId, clientId, ≥1 valid line item). safeParse before mutation; first issue surfaced as a toast with field path.

### ea8510df — Financial totals double-count discounts and optional items  COMPLEX
- category: data · severity: critical · files: estimates/page.tsx, invoices/page.tsx, create-estimate-modal.tsx
- Two bugs fused: (1) calculateLineTotal already bakes per-line discount into lineTotal, but the submit reducers also summed it into discountAmount and subtracted it again — every line discount applied twice. (2) Submit totals included optional+deselected items even though the on-screen total filtered them out.
- Submit reducers now mirror LineItemEditor's display reduce exactly: filter (isOptional && !isSelected), keep discountAmount at 0 (no doc-level discount input exists in these modals).

### ea4f7f2d — Product options cannot be used in estimates/invoices  ESCALATED
- category: functional · severity: high · file: src/components/ops/line-item-editor.tsx
- Cannot ground a selected-option payload on line items: the product-options authoring module is not present on main (only src/app/(dashboard)/products/page.tsx exists; no [id]/options route, no use-product-options hook, no ProductOptionsService). The product_options table is in database.types.ts but there is no shape on main for selected options (single vs multi-select, price modifier vs quantity multiplier) to snapshot. Per batch-03 instructions this overlaps with batch-01 escalated bug ae2cf28a — escalated rather than scaffolded.

## Test plan
- [ ] Visit /locked as admin/owner with expired sub → pricing grid; as non-admin with expired sub → request-reactivation CTA; as admin unseated → manage team CTA; as non-admin unseated → request access.
- [ ] Force /api/stripe/payment-methods to 500 → Billing tab shows error block with retry button.
- [ ] Cancel a subscription → status stays active until Stripe webhook fires customer.subscription.deleted.
- [ ] Trial company clicks Upgrade → /api/stripe/subscribe returns checkout URL (not 409).
- [ ] Return to /settings?result=success → loading toast → success only after status flips to active/grace + paid plan.
- [ ] Leave a session idle past token TTL, then click Sync Now / save in import wizard / refresh billing → no silent 401.
- [ ] Send an estimate or invoice → customer email lands; status flips to sent only after SendGrid succeeds.
- [ ] Try to save an estimate with no client / no line item → toast with field error; nothing persists.
- [ ] Add a line discount and an optional+deselected item → submit total matches the displayed total.

### DIFF
--- a/src/app/(auth)/locked/page.tsx
+++ b/src/app/(auth)/locked/page.tsx
@@ -1,13 +1,28 @@
 "use client";
 
-import { useCallback, useState } from "react";
-import { ShieldOff, Check, Headphones, Zap, Crown, Building2, Loader2 } from "lucide-react";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import {
+  ShieldOff,
+  UserX,
+  Check,
+  Headphones,
+  Zap,
+  Crown,
+  Building2,
+  Users,
+  Loader2,
+} from "lucide-react";
 import { cn } from "@/lib/utils/cn";
-import { TIER_CONFIG, type SubscriptionTier } from "@/lib/subscription";
+import {
+  getLockoutReason,
+  TIER_CONFIG,
+  type SubscriptionTier,
+} from "@/lib/subscription";
 import { Button } from "@/components/ui/button";
 import { useDictionary } from "@/i18n/client";
 import { OpsLockup } from "@/components/brand";
-import { useAuthStore } from "@/lib/store/auth-store";
+import { useAuthStore, selectIsAdminOrOwner } from "@/lib/store/auth-store";
+import { requireSupabase } from "@/lib/supabase/helpers";
 import { toast } from "sonner";
 
 // ─── Tier Visual Config ──────────────────────────────────────────────────────
@@ -44,7 +59,155 @@ const TIER_DISPLAY: Record<Exclude<SubscriptionTier, "trial">, {
   },
 };
 
-// ─── Pricing Card ────────────────────────────────────────────────────────────
+// ─── Admin Name Hook (mirrors LockoutOverlay.useAdminNames) ──────────────────
+
+function useAdminNames(adminIds: string[] | undefined) {
+  const [admins, setAdmins] = useState<{ id: string; name: string }[]>([]);
+
+  useEffect(() => {
+    if (!adminIds?.length) return;
+
+    let cancelled = false;
+
+    async function fetchAdminNames() {
+      try {
+        const supabase = requireSupabase();
+        const { data } = await supabase
+          .from("users")
+          .select("id, first_name, last_name")
+          .in("id", adminIds!);
+
+        if (cancelled || !data) return;
+
+        setAdmins(
+          data.map((u: { id: string; first_name: string; last_name: string }) => ({
+            id: u.id,
+            name: [u.first_name, u.last_name].filter(Boolean).join(" ") || "Admin",
+          }))
+        );
+      } catch {
+        // Silently fail — admin names are cosmetic
+      }
+    }
+
+    fetchAdminNames();
+    return () => { cancelled = true; };
+  }, [adminIds]);
+
+  return admins;
+}
+
+// ─── Cooldown helpers (mirrors LockoutOverlay) ───────────────────────────────
+
+const COOLDOWN_MS = 24 * 60 * 60 * 1000;
+
+function getCooldownKey(userId: string): string {
+  return `ops-lockout-request-${userId}`;
+}
+
+function isWithinCooldown(userId: string): boolean {
+  if (typeof window === "undefined") return false;
+  const stored = localStorage.getItem(getCooldownKey(userId));
+  if (!stored) return false;
+  try {
+    const { timestamp } = JSON.parse(stored);
+    return Date.now() - timestamp < COOLDOWN_MS;
+  } catch {
+    return false;
+  }
+}
+
+function setCooldown(userId: string, reason: "subscription_expired" | "unseated"): void {
+  if (typeof window === "undefined") return;
+  localStorage.setItem(
+    getCooldownKey(userId),
+    JSON.stringify({ timestamp: Date.now(), reason })
+  );
+}
+
+// ─── Request Button ──────────────────────────────────────────────────────────
+
+function RequestButton({
+  label,
+  sentLabel,
+  userId,
+  adminIds,
+  companyId,
+  userName,
+  reason,
+}: {
+  label: string;
+  sentLabel: string;
+  userId: string;
+  adminIds: string[];
+  companyId: string;
+  userName: string;
+  reason: "subscription_expired" | "unseated";
+}) {
+  const [sent, setSent] = useState(() => isWithinCooldown(userId));
+  const [sending, setSending] = useState(false);
+
+  const noAdmins = adminIds.length === 0;
+
+  const handleRequest = useCallback(async () => {
+    if (sent || sending || noAdmins) return;
+    setSending(true);
+
+    try {
+      const supabase = requireSupabase();
+      const isReactivation = reason === "subscription_expired";
+
+      const rows = adminIds.map((adminId) => ({
+        user_id: adminId,
+        company_id: companyId,
+        type: "role_needed" as const,
+        title: isReactivation ? "Reactivation Request" : "Access Request",
+        body: isReactivation
+          ? `${userName} is requesting subscription reactivation`
+          : `${userName} is requesting seat restoration`,
+        is_read: false,
+        persistent: true,
+        action_url: isReactivation ? "/settings?tab=subscription" : "/team",
+        action_label: isReactivation ? "Manage Subscription" : "Manage Team",
+      }));
+
+      const { error } = await supabase.from("notifications").insert(rows);
+      if (!error) {
+        setCooldown(userId, reason);
+        setSent(true);
+      }
+    } catch {
+      // Silently fail
+    } finally {
+      setSending(false);
+    }
+  }, [sent, sending, noAdmins, adminIds, companyId, userName, reason, userId]);
+
+  if (noAdmins) return null;
+
+  return (
+    <Button
+      variant="primary"
+      size="lg"
+      className="w-full"
+      onClick={handleRequest}
+      disabled={sent || sending}
+    >
+      {sent ? (
+        <span className="flex items-center gap-1">
+          <Check className="w-[16px] h-[16px]" />
+          {sentLabel}
+        </span>
+      ) : sending ? (
+        <span className="animate-pulse-live">{label}</span>
+      ) : (
+        label
+      )}
+    </Button>
+  );
+}
+
+// ─── Pricing Card (admin expired state) ──────────────────────────────────────
 
 function PricingCard({
   tier,
@@ -60,9 +223,7 @@ function PricingCard({
 
   // Initiate Stripe Checkout — webhook is the only writer of
   // `companies.subscription_status='active'`, so abandoning the Stripe-hosted
-  // checkout leaves the lockout in place. This is the same fix as the
-  // CompactPricingCard inside the dashboard lockout overlay (bug
-  // ac030a6c-6022-46dd-8d7b-5d477baaec53).
+  // checkout leaves the lockout in place.
   const handleSubscribe = useCallback(async () => {
     if (loading) return;
     if (!companyId) {
@@ -114,7 +275,6 @@ function PricingCard({
         display.popular && "ring-1 ring-ops-amber/20"
       )}
     >
-      {/* Popular badge */}
       {display.popular && (
         <div className="absolute -top-[12px] left-1/2 -translate-x-1/2">
           <span className="font-mono text-micro uppercase tracking-[0.2em] bg-ops-amber text-text-inverse px-1.5 py-0.5 rounded-sm">
@@ -123,7 +283,6 @@ function PricingCard({
         </div>
       )}
 
-      {/* Header */}
       <div className="flex items-center gap-1 mb-2">
         <div className={cn("p-1 rounded bg-fill-neutral-dim", display.accentClass)}>
           {display.icon}
@@ -133,7 +292,6 @@ function PricingCard({
         </div>
       </div>
 
-      {/* Price */}
       <div className="flex items-baseline gap-0.5 mb-2">
         <span className="font-mono text-[36px] leading-none text-text tracking-tight">
           ${config.price}
@@ -141,12 +299,10 @@ function PricingCard({
         <span className="font-mohave text-body-sm text-text-3">/mo</span>
       </div>
 
-      {/* Seat count */}
       <div className={cn("inline-flex items-center gap-0.5 px-1 py-0.5 rounded text-caption-sm font-mono mb-2 w-fit", display.badgeClass)}>
         {config.maxSeats} {t("locked.seatsIncluded")}
       </div>
 
-      {/* Features */}
       <ul className="flex flex-col gap-1 mb-3 flex-1">
         {config.features.map((feature) => (
           <li key={feature} className="flex items-start gap-1">
@@ -156,7 +312,6 @@ function PricingCard({
         ))}
       </ul>
 
-      {/* CTA */}
       <Button
         variant={display.popular ? "accent" : "default"}
         size="lg"
@@ -174,12 +329,108 @@ function PricingCard({
   );
 }
 
+// ─── Admin Display ───────────────────────────────────────────────────────────
+
+function AdminDisplay({
+  admins,
+}: {
+  admins: { id: string; name: string }[];
+}) {
+  const { t } = useDictionary("auth");
+
+  if (admins.length === 0) return null;
+
+  const primaryAdmin = admins[0];
+  const othersCount = admins.length - 1;
+
+  return (
+    <div className="flex items-center gap-1 mt-2 mb-3">
+      <span className="font-mono text-[11px] uppercase tracking-[0.15em] text-text-3">
+        {t("lockout.adminLabel")}:
+      </span>
+      <span className="font-mohave text-body text-text font-medium">
+        {primaryAdmin.name}
+      </span>
+      {othersCount > 0 && (
+        <span className="font-mohave text-body-sm text-text-3">
+          (+{othersCount} {t("lockout.adminOthers")})
+        </span>
+      )}
+    </div>
+  );
+}
+
+// ─── Footer ──────────────────────────────────────────────────────────────────
+
+function PageFooter({ sysMessageKey }: { sysMessageKey: string }) {
+  const { t } = useDictionary("auth");
+  return (
+    <div className="text-center space-y-1">
+      <p className="font-mohave text-body-sm text-text-3">
+        {t("locked.guarantee")}
+      </p>
+      <div className="flex items-center justify-center gap-2">
+        <a
+          href="mailto:support@opsapp.co"
+          className="inline-flex items-center gap-0.5 font-mohave text-body-sm text-text-2 hover:text-text underline underline-offset-4 transition-colors"
+        >
+          <Headphones className="w-[14px] h-[14px]" />
+          {t("locked.contactSupport")}
+        </a>
+        <span className="text-text-mute">|</span>
+        <a
+          href="/login"
+          className="font-mohave text-body-sm text-text-3 hover:text-text-2 underline underline-offset-4 transition-colors"
+        >
+          {t("locked.differentAccount")}
+        </a>
+      </div>
+      <p className="font-mono text-micro text-text-mute tracking-wider mt-2 opacity-40">
+        {t(sysMessageKey)}
+      </p>
+    </div>
+  );
+}
+
 // ─── Page ────────────────────────────────────────────────────────────────────
 
 export default function LockedPage() {
   const { t } = useDictionary("auth");
   const company = useAuthStore((s) => s.company);
+  const currentUser = useAuthStore((s) => s.currentUser);
+  const isAdmin = useAuthStore(selectIsAdminOrOwner);
+
   const companyId = company?.id;
+  const userId = currentUser?.id ?? null;
+  const userName = currentUser
+    ? [currentUser.firstName, currentUser.lastName].filter(Boolean).join(" ") || "A team member"
+    : "A team member";
+
+  // Compute actual lockout reason. Returns null if not locked at all,
+  // or "subscription_expired" / "unseated" otherwise.
+  const lockoutReason = useMemo(
+    () => getLockoutReason(company, userId),
+    [company, userId]
+  );
+
+  const admins = useAdminNames(company?.adminIds);
+
+  // Default state: no concrete reason yet (auth/company still loading, or no
+  // lockout at all). Fall back to the expired-admin pricing layout — same as
+  // the legacy behavior — so the page is never blank, but the heading is
+  // generic and we don't claim a specific reason we haven't proven.
+  const showAdminExpired =
+    lockoutReason === "subscription_expired" && isAdmin;
+  const showMemberExpired =
+    lockoutReason === "subscription_expired" && !isAdmin;
+  const showAdminUnseated = lockoutReason === "unseated" && isAdmin;
+  const showMemberUnseated = lockoutReason === "unseated" && !isAdmin;
+  const showFallbackPricing =
+    !showAdminExpired &&
+    !showMemberExpired &&
+    !showAdminUnseated &&
+    !showMemberUnseated;
+
   return (
     <div className="flex flex-col items-center min-h-screen px-2 py-5">
       {/* Logo */}
@@ -193,66 +444,159 @@ export default function LockedPage() {
         </p>
       </div>
 
-      {/* Status indicator */}
+      {/* Status icon — color depends on reason */}
       <div className="flex items-center gap-1 mb-2">
-        <div className="p-1 rounded-full bg-ops-error/15">
-          <ShieldOff className="w-[24px] h-[24px] text-ops-error" />
+        <div
+          className={cn(
+            "p-1 rounded-full",
+            lockoutReason === "unseated"
+              ? "bg-ops-amber/15"
+              : "bg-ops-error/15"
+          )}
+        >
+          {showAdminUnseated && (
+            <Users className="w-[24px] h-[24px] text-ops-amber" />
+          )}
+          {showMemberUnseated && (
+            <UserX className="w-[24px] h-[24px] text-ops-amber" />
+          )}
+          {(showAdminExpired || showMemberExpired || showFallbackPricing) && (
+            <ShieldOff className="w-[24px] h-[24px] text-ops-error" />
+          )}
         </div>
       </div>
 
-      {/* Heading */}
-      <div className="text-center mb-1 max-w-[600px]">
-        <h2 className="font-mohave text-display text-text mb-1">
-          {t("locked.title")}
-        </h2>
-        <p className="font-mohave text-body text-text-2 leading-relaxed">
-          {t("locked.description")}
-        </p>
-      </div>
+      {/* ── State A: Admin / Owner with expired subscription ── */}
+      {(showAdminExpired || showFallbackPricing) && (
+        <>
+          <div className="text-center mb-1 max-w-[600px]">
+            <h2 className="font-mohave text-display text-text mb-1">
+              {t(
+                showAdminExpired
+                  ? "lockout.expiredAdmin.title"
+                  : "locked.title"
+              )}
+            </h2>
+            <p className="font-mohave text-body text-text-2 leading-relaxed">
+              {t(
+                showAdminExpired
+                  ? "lockout.expiredAdmin.body"
+                  : "locked.description"
+              )}
+            </p>
+          </div>
 
-      {/* Divider */}
-      <div className="w-full max-w-[800px] flex items-center gap-2 my-3">
-        <div className="flex-1 h-px bg-border" />
-        <span className="font-mono text-[11px] uppercase tracking-[0.3em] text-text-3">
-          {t("locked.selectPlan")}
-        </span>
-        <div className="flex-1 h-px bg-border" />
-      </div>
+          <div className="w-full max-w-[800px] flex items-center gap-2 my-3">
+            <div className="flex-1 h-px bg-border" />
+            <span className="font-mono text-[11px] uppercase tracking-[0.3em] text-text-3">
+              {t("locked.selectPlan")}
+            </span>
+            <div className="flex-1 h-px bg-border" />
+          </div>
 
-      {/* Pricing grid */}
-      <div className="grid grid-cols-1 md:grid-cols-3 gap-2 w-full max-w-[900px] mb-4">
-        <PricingCard tier="starter" companyId={companyId} />
-        <PricingCard tier="team" companyId={companyId} />
-        <PricingCard tier="business" companyId={companyId} />
-      </div>
+          <div className="grid grid-cols-1 md:grid-cols-3 gap-2 w-full max-w-[900px] mb-4">
+            <PricingCard tier="starter" companyId={companyId} />
+            <PricingCard tier="team" companyId={companyId} />
+            <PricingCard tier="business" companyId={companyId} />
+          </div>
 
-      {/* Footer */}
-      <div className="text-center space-y-1">
-        <p className="font-mohave text-body-sm text-text-3">
-          {t("locked.guarantee")}
-        </p>
-        <div className="flex items-center justify-center gap-2">
-          <a
-            href="mailto:support@opsapp.co"
-            className="inline-flex items-center gap-0.5 font-mohave text-body-sm text-text-2 hover:text-text underline underline-offset-4 transition-colors"
-          >
-            <Headphones className="w-[14px] h-[14px]" />
-            {t("locked.contactSupport")}
-          </a>
-          <span className="text-text-mute">|</span>
-          <a
-            href="/login"
-            className="font-mohave text-body-sm text-text-3 hover:text-text-2 underline underline-offset-4 transition-colors"
-          >
-            {t("locked.differentAccount")}
-          </a>
+          <PageFooter
+            sysMessageKey={
+              showAdminExpired
+                ? "lockout.expiredAdmin.sysMessage"
+                : "locked.sysMessage"
+            }
+          />
+        </>
+      )}
+
+      {/* ── State B: Non-admin / non-owner with expired subscription ── */}
+      {showMemberExpired && (
+        <div className="w-full max-w-[520px] flex flex-col">
+          <div className="text-center mb-1">
+            <h2 className="font-mohave text-display text-text mb-1">
+              {t("lockout.expiredMember.title")}
+            </h2>
+            <p className="font-mohave text-body text-text-2 leading-relaxed">
+              {t("lockout.expiredMember.body")}
+            </p>
+          </div>
+
+          <AdminDisplay admins={admins} />
+
+          {userId && companyId && (
+            <RequestButton
+              label={t("lockout.expiredMember.requestReactivation")}
+              sentLabel={t("lockout.expiredMember.requestSent")}
+              userId={userId}
+              adminIds={company?.adminIds ?? []}
+              companyId={companyId}
+              userName={userName}
+              reason="subscription_expired"
+            />
+          )}
+
+          <div className="mt-4">
+            <PageFooter sysMessageKey="lockout.expiredMember.sysMessage" />
+          </div>
         </div>
+      )}
 
-        {/* System fingerprint for defense-tech aesthetic */}
-        <p className="font-mono text-micro text-text-mute tracking-wider mt-2 opacity-40">
-          {t("locked.sysMessage")}
-        </p>
-      </div>
+      {/* ── State C: Admin / Owner who is unseated ── */}
+      {showAdminUnseated && (
+        <div className="w-full max-w-[520px] flex flex-col">
+          <div className="text-center mb-1">
+            <h2 className="font-mohave text-display text-text mb-1">
+              {t("lockout.unseatedAdmin.title")}
+            </h2>
+            <p className="font-mohave text-body text-text-2 leading-relaxed">
+              {t("lockout.unseatedAdmin.body")}
+            </p>
+          </div>
+
+          <a href="/team" className="block mt-3">
+            <Button variant="primary" size="lg" className="w-full">
+              {t("lockout.unseatedAdmin.manageTeam")}
+            </Button>
+          </a>
+
+          <div className="mt-4">
+            <PageFooter sysMessageKey="lockout.unseatedAdmin.sysMessage" />
+          </div>
+        </div>
+      )}
+
+      {/* ── State D: Non-admin who is unseated ── */}
+      {showMemberUnseated && (
+        <div className="w-full max-w-[520px] flex flex-col">
+          <div className="text-center mb-1">
+            <h2 className="font-mohave text-display text-text mb-1">
+              {t("lockout.unseated.title")}
+            </h2>
+            <p className="font-mohave text-body text-text-2 leading-relaxed">
+              {t("lockout.unseated.body")}
+            </p>
+          </div>
+
+          <AdminDisplay admins={admins} />
+
+          {userId && companyId && (
+            <RequestButton
+              label={t("lockout.unseated.requestAccess")}
+              sentLabel={t("lockout.unseated.requestSent")}
+              userId={userId}
+              adminIds={company?.adminIds ?? []}
+              companyId={companyId}
+              userName={userName}
+              reason="unseated"
+            />
+          )}
+
+          <div className="mt-4">
+            <PageFooter sysMessageKey="lockout.unseated.sysMessage" />
+          </div>
+        </div>
+      )}
     </div>
   );
 }
--- a/src/app/(dashboard)/estimates/page.tsx
+++ b/src/app/(dashboard)/estimates/page.tsx
@@ -612,18 +612,35 @@ function EstimateFormModal({
       };
     });
 
+    // Total math — mirror LineItemEditor's display reduce EXACTLY so what
+    // the user sees on screen is what gets saved. (Bug ea8510df.)
+    //
+    // Two bugs were fused into the old reducer:
+    //   1. lineTotal is already discounted per-line (calculateLineTotal
+    //      multiplies by (1 - discountPercent/100)). The old code then also
+    //      computed sum(qty * price * pct/100) into discountAmount AND
+    //      subtracted that from total — so every line-level discount was
+    //      applied twice.
+    //   2. The reducer iterated every line, including optional items the
+    //      user deselected. The display filtered them out, so submit totals
+    //      were higher than what the customer saw.
+    //
+    // discountAmount stays at 0 here because the modal exposes no
+    // document-level discount input. If/when one is added it should be
+    // persisted from its own piece of state and subtracted ONCE.
     const totals = lineItems.reduce(
       (acc, li) => {
+        if (li.isOptional && !li.isSelected) return acc;
         const amt = computeAmount(li);
         return {
           subtotal: acc.subtotal + amt.lineTotal,
           taxAmount: acc.taxAmount + amt.tax,
-          discountAmount: acc.discountAmount + (li.discountPercent > 0 ? (li.quantity * li.unitPrice * li.discountPercent / 100) : 0),
         };
       },
-      { subtotal: 0, taxAmount: 0, discountAmount: 0 }
+      { subtotal: 0, taxAmount: 0 }
     );
-    const total = totals.subtotal + totals.taxAmount - totals.discountAmount;
+    const discountAmount = 0;
+    const total = Math.round((totals.subtotal + totals.taxAmount - discountAmount) * 100) / 100;
 
     const formData: Partial<CreateEstimate> & { companyId: string } = {
       companyId,
@@ -636,7 +653,7 @@ function EstimateFormModal({
       terms: termsAndConditions || null,
       subtotal: totals.subtotal,
       taxAmount: totals.taxAmount,
-      discountAmount: totals.discountAmount,
+      discountAmount,
       total,
       status: estimate?.status ?? EstimateStatus.Draft,
     };
--- a/src/app/(dashboard)/estimates/page.tsx
+++ b/src/app/(dashboard)/estimates/page.tsx
@@ -57,6 +57,7 @@ import {
 } from "@/lib/types/pipeline";
 import type { Estimate, Product, CreateEstimate, CreateLineItem } from "@/lib/types/pipeline";
 import { useAuthStore } from "@/lib/store/auth-store";
+import { createEstimateSchema } from "@/lib/schemas";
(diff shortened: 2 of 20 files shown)


src/app/(auth)/locked/page.tsx
```diff
@@ -193,66 +444,159 @@ export default function LockedPage() {
...
-            <Headphones className="w-[14px] h-[14px]" />
-            {t("locked.contactSupport")}
-          </a>
-          <span className="text-text-mute">|</span>
-          <a
-            href="/login"
-            className="font-mohave text-body-sm text-text-3 hover:text-text-2 underline underline-offset-4 transition-colors"
-          >
-            {t("locked.differentAccount")}
+          <PageFooter
```
jack-opsapp, line 503: Loading-state fallback note: when `lockoutReason === null` (auth/company still hydrating, or user is actually NOT locked at all), `showFallbackPricing` renders the admin expired pricing layout regardless of role. For a non-admin who is actually unseated, there is a brief flash of the wrong layout (admin pricing) until `company.adminIds` / `currentUser` populate. This matches the commit message intent ("never blank") and mirrors legacy behavior, so it is acceptable, but worth knowing — non-admins on slow networks may momentarily see Stripe CTAs they cannot complete. Not a blocker.

jack-opsapp:
REVIEW PASS [bug 59bfdae9] :: Locked page now mirrors LockoutOverlay's four-state matrix — calls getLockoutReason(company, userId), branches on selectIsAdminOrOwner, and renders the correct UI for admin+expired (pricing), member+expired (contact admin + request reactivation), admin+unseated (manage team CTA), and member+unseated (contact admin + request access). Cooldown logic, notification rows, RequestButton, and useAdminNames are byte-identical patterns to LockoutOverlay — good consistency. Translation keys reuse existing lockout.* namespace (verified in en/auth.json). One non-blocking note: when lockoutReason is null (auth still hydrating, or user routed to /locked while not actually locked), fallback renders admin pricing layout regardless of role — brief flash for non-admins on slow networks. Acceptable per commit intent (never blank). STATUS :: ALLOW MERGE.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -114,12 +114,21 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+    // was the bug (9f55103f): SubscriptionTab sent trial users to /subscribe
+    // and the route refused them with 409, blocking conversion entirely.
+    //
+    // 'grace' is a real paid subscription in past_due/paused state. Keep
+    // rejecting it — they need to update payment, not create a second sub.
     if (
       company.subscription_status &&
-      ["active", "trial", "grace"].includes(company.subscription_status)
+      ["active", "grace"].includes(company.subscription_status)
     ) {
```
jack-opsapp, line 132: Status state machine after this fix — allowed statuses passing the guard are: null, trial, expired, cancelled. Rejected: active, grace. That is correct for the trial-to-paid case and the re-subscribe case.

However there is a subtle hole that this PR does not close: when a trial user successfully POSTs /subscribe with a card that needs 3DS, Stripe returns status='incomplete'. mapStripeStatus returns null for 'incomplete' (subscription-mapping.ts line 50-52), so this route writes subscription_plan/subscription_period/max_seats but leaves subscription_status='trial'.

If that same user clicks Upgrade again before the webhook fires (3DS popup closed, network flake, retry on a different plan/period), they pass this N2 guard again because status is still 'trial'. The Stripe idempotency-key on the create call (line 257, scoped to company+plan+period, 24h TTL) protects against the SAME plan/period being submitted twice — but a different plan or period (e.g. user picks Starter, dismisses 3DS, then picks Team) generates a different idempotency key and creates a real second Stripe subscription.

Recommendation for a follow-up: also reject when there is a non-null subscription_ids_json with an in-flight 'incomplete' sub, OR move to a single 'pending_payment' status that this guard also rejects. Not blocking on this PR — flagging the gap.

src/app/api/stripe/cancel/route.ts
```diff
@@ -95,22 +95,39 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+    // (and ONLY that webhook) is the authoritative writer of
+    // subscription_status='cancelled' on companies.
+    //
+    // We deliberately do NOT write subscription_status here. Writing it as
+    // 'cancelled' immediately would lock the customer out of the app the
+    // instant they scheduled cancellation, even though they have already
+    // paid for the remaining days in the period. (Bug e67562e5.)
+    const updated = await stripe.subscriptions.update(subscriptionId, {
       cancel_at_period_end: true,
     });
```
jack-opsapp, line 110: [bug e67562e5] Fix verified correct. Stripe semantics confirmed: when `cancel_at_period_end=true`, Stripe leaves `subscription.status='active'` until the period ends, then fires `customer.subscription.deleted`. The webhook at `src/app/api/webhooks/stripe/route.ts:400-473` is the authoritative writer of `subscription_status='cancelled'`. The intermediate `customer.subscription.updated` event will fire with status "active" and `mapStripeStatus` (subscription-mapping.ts:36) returns "active" — confirmed no race that would re-lock the user. Comment on lines 98-107 is accurate and load-bearing.

src/app/api/stripe/cancel/route.ts
```diff
@@ -95,22 +95,39 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+    // fall back to current_period_end if Stripe hasn't populated cancel_at yet.
+    const cancelAtUnix =
+      updated.cancel_at ??
+      (updated as unknown as { current_period_end?: number | null })
+        .current_period_end ??
+      null;
+    const cancelAtIso =
+      cancelAtUnix !== null && cancelAtUnix !== undefined
+        ? new Date(cancelAtUnix * 1000).toISOString()
+        : null;
```
jack-opsapp, line 124: [bug e67562e5] `cancelAt` derivation is correct. Note: on Stripe API 2024-06+, `current_period_end` moved from the Subscription to the SubscriptionItem (see webhook.ts:346 which uses `subscription.items.data[0].current_period_end`). The unsafe cast on line 118 papers over the type but the fallback may resolve to `undefined` on newer API versions — `updated.cancel_at` will normally be set after scheduling a period-end cancel, so the fallback is rarely hit. Acceptable as-is; ideally future hardening uses `updated.items.data[0]?.current_period_end` to match the webhook. Not blocking.

src/app/api/stripe/subscribe/route.ts
```diff
@@ -114,12 +114,21 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
+    //
+    // IMPORTANT: 'trial' is NOT a Stripe subscription — it's the free in-app
+    // trial granted on signup. Trial companies have no Stripe subscription
+    // yet and MUST be allowed to convert to a paid plan. Rejecting them here
+    // was the bug (9f55103f): SubscriptionTab sent trial users to /subscribe
+    // and the route refused them with 409, blocking conversion entirely.
+    //
+    // 'grace' is a real paid subscription in past_due/paused state. Keep
+    // rejecting it — they need to update payment, not create a second sub.
     if (
```
jack-opsapp, line 129: Pre-existing security issue, not introduced here, but flagging on this PR because the trial-to-paid path is now actually reachable through the same code: decodeFirebaseToken at the top of this file (line 23-35) does NOT verify the JWT signature — it just base64-decodes the middle segment of the bearer token and reads payload.sub and payload.email.

Any client can forge a token with arbitrary sub/email and send it in the Authorization header. The route then looks up the company strictly by req.body.companyId, not by the authenticated user's actual company membership. So a malicious caller who knows another company's UUID could trigger Stripe customer create + subscription create on that company's record, with whatever plan they choose.

The Stripe customer create uses user.email from the forged token (line 161), so the attack would be visible after the fact, but the company row gets subscription_plan/max_seats overwritten and the legitimate customer's billing surface is corrupted.

Needs a real Firebase Admin SDK verifyIdToken call (verify signature + audience + exp), plus a membership check that the authenticated user belongs to the companyId being subscribed.

Out of scope for this hotfix — but this fix turns a previously inert code path into a live one. Worth a separate critical issue.

jack-opsapp:
REVIEW PASS [bug e67562e5] :: APPROVED — fix is correct.

**Verdict:** Removal of the immediate `subscription_status='cancelled'` write is exactly right. Cancel-at-period-end semantics now behave as the bible specifies.

**Wargame results:**
- **Stripe semantics:** Verified. Stripe holds `subscription.status='active'` until period boundary, then fires `customer.subscription.deleted`. Webhook at `src/app/api/webhooks/stripe/route.ts:400-473` is the authoritative writer. ✓
- **Webhook race:** Safe. The interim `customer.subscription.updated` event Stripe fires post-cancel-schedule carries status='active', which `mapStripeStatus` (`src/lib/stripe/subscription-mapping.ts:36`) maps to 'active' — no fight. ✓
- **Idempotency:** Safe. Double-cancel is a no-op against Stripe (`cancel_at_period_end: true` is idempotent) and route writes no local state. ✓
- **Lockout UX:** Status stays 'active' — user retains full access until period end. Returned `cancelAt` ISO date enables a 'Cancels on <date>' soft warning in the UI. Note: caller `useCancelSubscription` (`src/lib/hooks/use-company.ts:169`) discards the response body. Wiring the date into settings UI is a follow-up, not a blocker for this bug. ✓
- **Reactivate-after-cancel:** Unaffected by this commit. Webhook layer handles reactivation via `customer.subscription.updated`. ✓

**Pre-existing concerns flagged (NOT introduced by this fix, NOT blocking):**
- `decodeFirebaseToken` only base64-decodes; signature/expiry not verified.
- No role check — any authenticated caller with a known `companyId` can cancel.
- Recommend separate bug for hardening this route.

**Minor (not blocking):** Stripe API 2024-06+ moved `current_period_end` to the SubscriptionItem. Fallback on line 118 may resolve to undefined on newer API versions, but `updated.cancel_at` is reliably populated after scheduling — fallback rarely hits in practice.

STATUS: APPROVED

src/app/api/stripe/subscribe/route.ts
```diff
@@ -114,12 +114,21 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
...
       company.subscription_status &&
-      ["active", "trial", "grace"].includes(company.subscription_status)
+      ["active", "grace"].includes(company.subscription_status)
```
jack-opsapp, line 131: Idempotency analysis — the key is 'company-{companyId}-sub-{plan}-{period}' with Stripe's 24h TTL (line 257). For the trial-to-paid path this is correct: same plan+period submitted twice within 24h returns the same subscription. Good.

Two edge cases worth noting:

1. Re-subscribe after cancellation within 24h: a user who cancels and then re-subscribes to the SAME plan+period within the 24h window will hit the idempotency cache and get the OLD (cancelled) subscription back. Stripe will return the prior sub, this route writes subscription_ids_json with the old id, and the user thinks they're subscribed but no new charge happens. Low likelihood — most cancellation flows include a cooldown — but the failure mode is silent.

2. Plan switch during pending 3DS: see my comment on line 132. The N2 guard doesn't catch this because the row is still in 'trial' status while Stripe is 'incomplete'. Different plan+period means different idempotency key means a real second subscription.

Both belong in a dedicated upgrade/downgrade/reactivate route, which the comment on line 119 already acknowledges is not built yet.

src/app/(dashboard)/estimates/page.tsx
```diff
@@ -560,6 +561,39 @@ function EstimateFormModal({
...
+        description: null,
+        quantity: li.quantity,
+        unit: li.unit,
+        unitPrice: li.unitPrice,
+        isTaxable: li.isTaxable,
+        discountPercent: li.discountPercent,
+        productId: li.productId ?? null,
+        isOptional: li.isOptional,
+        isSelected: li.isSelected,
+      })),
```
jack-opsapp, line 586: Deselected optional line items still go through validation. Every entry in `lineItems` is checked against `lineItemFormSchema` (`name.min(1)`, `quantity.min(0.0001)`), but the totals reducer below at L631-641 silently skips `isOptional && !isSelected` rows because they have no customer impact. So a user who added a placeholder optional row, blanked its name, then deselected it gets blocked at save even though that row contributes nothing to the document. Either filter `isOptional && !isSelected` rows out of the payload before `safeParse`, or relax `lineItemFormSchema` for deselected lines.

src/app/(dashboard)/estimates/page.tsx
```diff
@@ -560,6 +561,39 @@ function EstimateFormModal({
...
+    });
```
jack-opsapp, line 587: Commit message claims "zero totals" no longer save, but neither `createEstimateSchema` nor `lineItemFormSchema` enforces a positive `unitPrice` or total. A line with `name="x"`, `quantity=1`, `unitPrice=0` passes `safeParse` cleanly, producing a $0 estimate. Schema currently requires only: name present, quantity > 0.0001, client present, ≥1 line item. If a $0 document is invalid by intent, add `unitPrice.min(0.01)` to `lineItemFormSchema` or a `.refine` on `createEstimateSchema` checking subtotal > 0. Same applies to `createInvoiceSchema` — the more dangerous case, since a $0 invoice is meaningless and likely a user error.

jack-opsapp:
**Review pass — bug 9f55103f — Stripe subscribe trial-to-paid (commit 16fcd3ef)**

**Verdict: APPROVE the fix as-is.** It correctly unblocks the trial-to-paid conversion path. The reasoning in the commit message is sound: 'trial' is the in-app free trial, NOT a Stripe trialing sub, so the N2 'already-subscribed' guard had no business rejecting it. Dropping 'trial' from the rejected set restores conversion. 'active' and 'grace' stay rejected — correct.

**Verified end-to-end:**
- SubscriptionTab POSTs companyId/plan/period to /api/stripe/subscribe (subscription-tab.tsx line 97-108). 402 handler routes user to billing to add a card — clean.
- Status state machine post-fix: null/trial/expired/cancelled pass; active/grace reject. Correct for trial-to-paid AND cancelled-then-re-subscribe.
- Trial-to-paid creates a NEW Stripe subscription via stripe.subscriptions.create (line 250) — correct because trial users have no prior Stripe sub. Not subscription.update.
- Stripe customer create has idempotency-key 'company-{id}-customer' + a compare-and-swap on the Supabase update (line 168-195) — concurrent /subscribe calls cannot orphan a Stripe customer.
- Subscription create has idempotency-key 'company-{id}-sub-{plan}-{period}' with Stripe's 24h TTL (line 257) — double-tap on the SAME plan/period returns the same sub.
- No-payment-method path is clean: 402 with code 'payment_method_required' (line 228-237).
- Post-subscribe DB state: subscription_plan + max_seats + subscription_end + subscription_ids_json get written immediately; subscription_status is derived from mapStripeStatus (null for 'incomplete', so left untouched until webhook). Webhook is the source of truth for status — correct.

**Three issues flagged in line comments — none block this PR, all worth follow-up:**

1. **(line 132) Plan-switch hole during pending 3DS.** N2 guard keys off subscription_status, but a trial user mid-3DS still has status='trial'. A different plan/period generates a different idempotency key and would create a real second Stripe subscription. Idempotency only protects same plan+period within 24h.

2. **(line 129) Pre-existing critical auth flaw.** decodeFirebaseToken does NOT verify the JWT signature — anyone can forge a token, pass any companyId, and trigger Stripe customer/subscription create against another company's record. This fix turns a previously inert code path into a live one, so it deserves urgent attention even though it predates this commit. Needs real Firebase Admin SDK verifyIdToken + a company-membership check.

3. **(line 131) Re-subscribe-after-cancellation within 24h.** Same plan+period within Stripe's idempotency TTL returns the cancelled sub silently. Low likelihood, silent failure mode.

**Race with webhook:** A late webhook from the prior trial-onboarding flow could race the subscribe write, but trial onboarding doesn't create Stripe subs (that's the whole point of this fix), so there is no late webhook from a trial. For the cancelled→re-subscribe case, the prior sub's webhook events would target the OLD subscription id, and the new write replaces subscription_ids_json with the new id, so a stale 'customer.subscription.deleted' on the old id would not match — webhook handler should ignore it. Safe.

Recommend merging. File separate critical issue for the auth flaw immediately.

src/app/(dashboard)/estimates/page.tsx
```diff
@@ -560,6 +561,39 @@ function EstimateFormModal({
...
+    if (!validation.success) {
+      const first = validation.error.issues[0];
+      const path = first?.path?.join(".") ?? "form";
+      toast.error("Can't save estimate", {
+        description: `${path}: ${first?.message ?? "Invalid input"}`,
+      });
```
jack-opsapp, line 593: Error UX: only the first issue is surfaced as a toast with the raw Zod path (e.g. `lineItems.2.name: Name is required`). For multi-error forms the user gets a single cryptic message and has to fix issues one at a time, re-submitting between each. Better: iterate `validation.error.issues`, group by top-level field, and either (a) show all of them in the toast description or (b) attach per-field error state to the form so each row/field highlights inline. The path string is also raw — `lineItems.0.quantity` is not what the user calls these. A small `pathToLabel` map (`clientId → Client`, `lineItems.N.name → Line N name`) would make these readable.

src/app/(dashboard)/invoices/page.tsx
```diff
@@ -571,6 +572,39 @@ function InvoiceFormModal({
...
+        quantity: li.quantity,
+        unit: li.unit,
+        unitPrice: li.unitPrice,
+        isTaxable: li.isTaxable,
+        discountPercent: li.discountPercent,
+        productId: li.productId ?? null,
+        isOptional: li.isOptional,
+        isSelected: li.isSelected,
+      })),
+    });
```
jack-opsapp, line 598: Same "deselected optional still validates" problem as the estimates side. Plus, `dueDate` here is required by `createInvoiceSchema` (`z.date()` — no `.optional()`), and the payload sends `undefined` if the user blanked the date field. That is fine in practice because the `useEffect` at L519 auto-computes `dueDate` from `issueDate` + `paymentTerms`, so the field is always populated unless the user manually clears it AFTER the effect runs. If they do, schema rejects with `dueDate: Required` — that is the correct outcome but worth noting that the displayed field has no required marker and no auto-recompute on blur.

jack-opsapp:
## Review pass — bug ce6495b9 (Estimates / Invoices submit validation)

**Verdict:** APPROVE WITH FINDINGS. The schema gate is correctly placed before any mutation in both `EstimateFormModal.handleSubmit` and `InvoiceFormModal.handleSubmit`, and it catches the headline cases (missing `clientId`, missing `companyId`, blank line item `name`, zero/negative `quantity`, empty `lineItems` array). All four are confirmed blocked because `createEstimateSchema` / `createInvoiceSchema` / `lineItemFormSchema` enforce those constraints.

### What is now blocked
- Save with no client selected → `clientId: Client is required`
- Save with zero line items → `lineItems: At least one line item is required`
- Save with a line item missing a name → `lineItems.N.name: Name is required`
- Save with `quantity` ≤ 0.0001 → `lineItems.N.quantity: Quantity must be positive`
- Invoice save without `dueDate` → `dueDate: Required` (rare; auto-computed by effect)

### What is still allowed (worth knowing)
1. **$0 documents pass.** Neither schema enforces `unitPrice > 0` or `total > 0`. A single line `{ name: "x", quantity: 1, unitPrice: 0 }` saves cleanly. Commit message claim of "zero totals blocked" is inaccurate — only zero *quantity* is blocked. If $0 financial docs are invalid by intent, add `unitPrice.min(0.01)` to `lineItemFormSchema` or a `.refine(total > 0)` on the document schemas.
2. **Cross-tenant `clientId`.** Schema only checks `clientId.min(1)`. A spoofed UUID from another tenant passes the form gate; only DB-level isolation rules (RLS) stop the actual write. Not introduced by this commit; flagging as latent.
3. **Deselected optional line items still validate.** Totals reducer (L631-641 estimates, L626-636 invoices) correctly excludes `isOptional && !isSelected` rows, but the validation payload includes them. A blank/deselected optional row will block save even though it has no customer impact and contributes nothing to the document. Fix: filter `isOptional && !isSelected` out of the `safeParse` payload, or filter them out of the persisted `mappedLineItems` entirely.

### Error UX gap
Only first issue is surfaced via toast as raw Zod path (`lineItems.2.name: Name is required`). No inline field highlighting, no human labels. Acceptable for unblocking the bug; should be tracked as follow-up.

### Untouched by this commit
- Status / draft handling: `status` is not in either form schema and falls through unmolested. There is no "save as draft with looser rules" path — draft saves go through the same gate. Pre-existing bad rows in the DB load into the form fine (validation runs at submit, not at mount).
- Permission gating: page-level (`can("estimates.create" / "estimates.edit")`) controls who can open the modal; this commit does not change that surface.

Overall: ships, but I'd push a follow-up for items (1) and (3) before treating this bug fully closed.

jack-opsapp:
## Review pass — bug f920326c (billing recovery success-before-current)

**Verdict:** approve with required follow-up. The fix correctly addresses the original symptom (no more "Subscription active" toast while lockout overlay is still up), but a critical effect-rerun race undermines the polling guarantee in production builds. Several smaller issues should be cleaned up in a follow-up commit before merge.

### Wargame summary

| Concern | Status |
|---|---|
| Webhook timing (no false success during lag) | Handled — 2s poll up to 30s, soft "still finalizing" on timeout |
| Stale auth / JWT claims | Not a vector here — lockout reads from Zustand `company.subscriptionStatus`; auth is Firebase, no Supabase JWT subscription claims. Pushing fresh company via `setAuthCompany` re-evaluates immediately. |
| Polling backoff | Acceptable (fixed 2s, max 15 attempts) — no exponential needed for a 30s budget |
| Webhook-lands-during-poll race | Handled — Realtime listener in `lockout-overlay.tsx` (line 91, `useRealtimeCompanyUpdates`) is a redundant safety net |
| Sign-out mid-recovery | **Gap** — in-flight `refetch` is not aborted; `setAuthCompany(fresh)` can repopulate `company` after logout cleared it. Brief stale-company window. |
| Failure path (`past_due`, declined) | **Insufficient** — neutral toast with no CTA; user is stuck on the page with no retry path |

### Critical issues (must address)
1. **Effect rerun race** (line 444): `router.replace` re-emits `searchParams` → effect cleanup fires while the first poll is still in-flight, killing the loading toast. See inline comment.
2. **Plan allow-list** (line 391): hard-coded Starter/Team/Business. A new tier silently fails to entitle. See inline comment.

### Should-fix
3. `setTimeout` handle not cleared on cleanup (line 435)
4. No `AbortController` on `refetch` — sign-out can write stale company to auth store
5. Failure-path UX has no retry CTA (line 429)
6. Variable shadowing of `result` (line 406) — lint/readability

Five inline comments on commit `0610bcaa`.

src/lib/hooks/use-email-connections.ts
```diff
@@ -76,7 +77,7 @@ export function useTriggerEmailSync() {
 
   return useMutation({
     mutationFn: async (connectionId: string) => {
-      const response = await fetch("/api/integrations/email/manual-sync", {
+      const response = await authedFetch("/api/integrations/email/manual-sync", {
```
jack-opsapp, line 80: Direct twin not migrated: `src/lib/hooks/use-gmail-connections.ts:73` (`useTriggerGmailSync`) has the exact same pattern — POST to `/api/integrations/gmail/manual-sync` with plain `fetch` — and is invoked from the same Settings > Integrations surface as this hook. Same long-running UI, same token-stale risk, same workflow-critical action (manual mailbox sync). Also `src/lib/hooks/use-gmail-import.ts:58` (`startImport`) for the Gmail historical import flow. Both should adopt `authedFetch` in this PR — otherwise the fix is partial: IMAP users get refresh, Gmail users still hit silent failure.

src/lib/hooks/use-billing.ts
```diff
@@ -45,7 +46,7 @@ export function usePaymentMethods() {
   return useQuery({
     queryKey: billingKeys.paymentMethods(companyId),
     queryFn: async (): Promise<PaymentMethod[]> => {
-      const res = await fetch(`/api/stripe/payment-methods?companyId=${companyId}`);
+      const res = await authedFetch(`/api/stripe/payment-methods?companyId=${companyId}`);
```
jack-opsapp, line 49: Token-refresh-failure UX: when the underlying `authedFetch` retry yields another 401 or `getIdToken(true)` returns null (session revoked, signed-out, network failure), it just returns the 401 Response. Here that becomes `throw new Error('Failed to fetch payment methods')` → red TanStack-Query toast. The user never gets a clean re-auth prompt. Same story in `useStripeInvoices` (line 65) and `useRemovePaymentMethod` (line 90). For a billing surface this matters — operators will think their card data failed to load when in fact their session expired. Recommend a follow-up that wires the unrecoverable-401 case into `auth-store.signOut()` + redirect to `/login?reason=expired`. Not blocking this PR.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -154,7 +154,7 @@ export function ImportPipelineWizard({
     if (fingerprint === lastSaveFingerprintRef.current) return;
 
     try {
-      await fetch("/api/integrations/email/connection", {
+      await authedFetch("/api/integrations/email/connection", {
```
jack-opsapp, line 157: Adjacent unmigrated workflow: `src/components/settings/email-setup-wizard.tsx:283` (`restoreScanFromJob`) and `:583` (`scan-start`) are the *exact* same long-lived wizard pattern this PR is fixing — a settings page that can sit open for many minutes while the user navigates a multi-step setup. If the IMAP/generic-email wizard needs `authedFetch` to recover from stale tokens during save/import/activate, so does the Gmail-specific wizard. Recommend extending the migration to those two call sites in the same PR, since they share the failure mode and the QA repro.

src/components/settings/import-pipeline-wizard.tsx
```diff
@@ -788,7 +788,7 @@ export function ImportPipelineWizard({
         },
       };
 
-      const res = await fetch("/api/integrations/email/import", {
+      const res = await authedFetch("/api/integrations/email/import", {
```
jack-opsapp, line 791: Critical workflow-action surfaces not covered by this PR (same risk profile, same suspected-file class):\n\n- `src/components/ops/lockout-overlay.tsx:292` — `/api/stripe/checkout` POST. Subscription-recovery flow that an idle session is *very* likely to land in (lockout screens appear after the session has been sitting open).\n- `src/components/settings/subscription-tab.tsx:97` — `/api/stripe/subscribe` POST. Long-lived settings page, billing-critical.\n- `src/components/settings/addons-section.tsx:309` — `/api/stripe/billing-portal` POST. Same.\n- `src/lib/hooks/use-addons.ts:157/181/322` — three Stripe add-on calls.\n- `src/lib/api/services/company-service.ts:219/248/272` — `setup-intent`, `subscribe`, `cancel` Stripe endpoints all bare-fetch.\n- `src/components/settings/data-privacy-tab.tsx:28/70` — data export, account deletion. Workflow-critical, irreversible.\n- `src/lib/hooks/use-users.ts:252` — admin password reset.\n- `src/components/agent/confirm-schedule-button.tsx:94/129` — schedule confirmation, customer-facing.\n\nThe inbox layer has a separate problem: `src/lib/hooks/use-inbox-threads.ts` uses its own `authHeaders()` helper (line 208) that attaches the Bearer token but does *not* refresh on 401. Every inbox read/write goes through that helper and inherits the same bug. Either migrate those to `authedFetch` too or have `authHeaders()` use the same 401-retry path.

jack-opsapp:
**Review Pass — bug 728027bb** | verdict: **conditional approve — partial fix, blast radius incomplete**

**What the fix does correctly:**
- Converts the 7 call sites named in QA (`use-billing` × 3, `use-email-connections` × 1, `import-pipeline-wizard` × 4) to `authedFetch`.
- `authedFetch` itself looks correct: header merge handles `Headers` / array / record forms without clobbering, retry only on 401 (won't accidentally fire on 5xx), uses `getIdToken(true)` to force-refresh from Firebase. No Content-Type clobbering. `init` is spread first so headers override cleanly.
- No external API leakage in the diff — only same-origin `/api/*` paths converted, so Stripe/SendGrid/Bubble direct calls aren't at risk of receiving a Firebase JWT.
- Body shapes (`JSON.stringify`) and response handling unchanged — no compat break with TanStack `onError`.

**What this PR misses (wide blast radius — bug is about the *pattern*, not just these 7 lines):**
1. **Gmail twins of the migrated IMAP hooks** — same workflow, same risk, not touched:
   - `use-gmail-connections.ts:73` (`useTriggerGmailSync`)
   - `use-gmail-import.ts:58` (`startImport`)
   - `email-setup-wizard.tsx:283/583` (Gmail wizard scan-start / restore — direct parallel to `import-pipeline-wizard` save/import/activate)
2. **Billing surfaces still on bare fetch** — same idle-session failure mode as the use-billing hooks that *were* fixed:
   - `lockout-overlay.tsx:292` (`/api/stripe/checkout` — recovery flow specifically reached *from* an idle session)
   - `subscription-tab.tsx:97`, `addons-section.tsx:309`, `use-addons.ts` × 3, `company-service.ts:219/248/272`
3. **Inbox uses a parallel helper** `authHeaders()` in `use-inbox-threads.ts:208` that attaches a Bearer token but *does not refresh on 401*. Every inbox read/write inherits the bug. Bug ID lists `/inbox` in the URL field — this is the primary user-visible inbox surface and it was not migrated.
4. **Irreversible/sensitive actions** still on bare fetch: `data-privacy-tab.tsx:28/70` (export, account-delete), `use-users.ts:252` (admin password reset).

**Unrecoverable-401 UX gap (not introduced by this PR, but cited):** when refresh fails or retry yields another 401, the wrapper returns the 401 to the caller. Every migrated call site converts that to a generic 'Failed to fetch…' toast — the bug summary calls out *silent failure*, and this is still semi-silent. Recommend a follow-up that routes unrecoverable-401 → `auth-store.signOut()` + `/login?reason=expired`.

**Other wargame checks:**
- *Headers / FormData / streams:* no converted site uses FormData, raw streams, or non-JSON bodies — safe.
- *SSR:* none of the converted sites run server-side — wrapper relies on `auth.currentUser` so it's correctly client-only.
- *External APIs:* none accidentally converted.
- *TanStack `onError`:* error shape unchanged (`throw new Error(...)`) — preserved.

**Verdict:** Merge as a *first pass* with a follow-up issue tracking (1) the Gmail twin migrations, (2) the billing/lockout/data-privacy surfaces, (3) the inbox `authHeaders()` parallel-pattern, (4) the unrecoverable-401 redirect-to-login UX. PR title says 'critical billing and email workflows' which is accurate for what it does, but the bug title 'Critical workflows bypass token refresh helper' implies a broader sweep that this PR doesn't deliver.

— bug-triage agent (review pass for `728027bb`)

jack-opsapp (commented):
**Review pass — bug a0bd0021 (critical, send-money flow).** Wiring is correct in shape: routes call SendGrid before stamping `sent_at`, and the email-first ordering is right. However several issues remain on this new code path that should be addressed before this can be considered safe for production sends. Severity ordered, most-critical first:

**1. (BLOCKER) No cross-tenant guard.** `verifyAdminAuth` only confirms the caller has *some* valid Supabase/Firebase token (see `src/lib/firebase/admin-verify.ts:151-166` — returns `VerifiedUser` from JWT signature only, no company/role check). Both new routes then load the estimate/invoice by id alone and dispatch SendGrid using the document's `company_id`. Any authenticated user in any tenant can POST `/api/estimates/{anyone's-id}/send` and trigger an email from another company's portal sender to the victim company's client. This is a tenancy break on a paid-tier transactional channel. The route MUST resolve the caller's company (e.g. via `user_companies` / membership lookup keyed on `admin.uid`) and assert `caller_company_id === estimate.company_id` before proceeding.

**2. (BLOCKER) Not idempotent — double-clicking Send fires a second email.** Neither route checks current `status`. A user who double-clicks, retries on slow network, or hits the endpoint twice from two tabs will send the customer two emails, mint two portal tokens, and update `sent_at` twice. For a money-bearing channel this is a customer-trust failure. Add an early guard: if `estimate.status === 'sent'` (or the equivalent invoice status), 409 with a clear message and let the UI offer a 'Resend' affirmative confirmation.

**3. (BLOCKER) Email-sent / DB-write inconsistency leaves estimate in a lying state.** When SendGrid succeeds but the `update({status:'sent', sent_at})` write fails, the route logs and returns `200 {success:true, sent_at}`. The UI then optimistically invalidates and refetches, which shows `sent_at: null` and `status: <prior>` — operator believes the send failed and clicks again, sending a second email (compounds issue #2). Either retry the DB write, queue a reconciliation job, or persist a separate `document_sends` audit row first and let `status` be derived. At minimum, surface a distinct error code to the UI so the operator knows 'email delivered, status stamping failed' and does NOT re-send.

**4. (HIGH) Recipient email regex permits multi-address injection.** `/^[^\s@]+@[^\s@]+\.[^\s@]+$/` does not reject commas, semicolons, or angle brackets. `a@b.com,c@evil.com` or `victim@x.com>,<attacker@y.com` will pass and be handed to SendGrid's `to:` field. Depending on `gatedSend` internals this can either fail at SendGrid or be split into a multi-recipient send. Tighten to a strict single-address parser, reject any address containing `,`, `;`, `<`, `>`, or whitespace, and normalize to lowercase before passing to `createPortalToken` AND `sendEstimateReady` so the portal token's bound email matches the recipient header exactly.

**5. (HIGH) No audit trail of who sent what to whom.** Stamping `sent_at` on the estimate/invoice row records *that* it was sent, not *who* clicked Send, *what* recipient was used (especially when overridden), *which* portal token was minted, or *whether* SendGrid accepted it. For a billing-adjacent channel this is the audit you want when a customer disputes 'I never got the invoice.' Recommend inserting a `document_sends` (or analogous) row with `actor_user_id`, `recipient_email`, `portal_token_id`, `sendgrid_message_id` (from gatedSend response if surfaced), and `sent_at` — and treating it as the source of truth, with the column on the parent row derived.

**6. (MEDIUM) Portal token leakage via override-email path.** `createPortalToken(companyId, clientId, recipient)` binds the magic-link token to the typed-in email. The operator can type ANY address into the Send dialog (the input is free-text — see `send-estimate-flow.tsx:107-112`), so a careless or malicious operator can mint a portal token bound to an arbitrary address and grant that address access to `/portal/estimates/{id}` for this client. This is by design for 'send to a different contact' but it has no second-confirmation when the typed address differs from `client.email`, and no logging. Add: (a) a confirmation step in the dialog when `email !== client.email`, and (b) an audit log entry capturing the override.

**7. (MEDIUM) Reply-to goes to the OPS portal address, not the sending company.** `sendEstimateReady` / `sendInvoiceReady` set `replyTo: getPortalFromEmail()`. When a customer hits Reply to a quote from `${companyName}`, their reply goes to OPS, not to the contractor. For a Send Estimate flow this is almost certainly wrong product behavior — the contractor is the one who needs to hear back. Confirm intent; if reply-to should be the company's contact email, the route should resolve `company.contact_email` (or `company.owner.email`) and pass it explicitly.

**8. (MEDIUM) No SendGrid sandbox / non-prod isolation visible in this path.** `gatedSend` is invoked unconditionally — there is no environment guard in the route to prevent preview / staging deployments from emailing real customers if a tester runs Send against a production-cloned dataset. Confirm `gatedSend` itself short-circuits when `NODE_ENV !== 'production'` or when a `SENDGRID_SANDBOX` flag is set; if it does not, the route should refuse to send when running on preview.

**9. (LOW) Magic-link URL silently degrades when `NEXT_PUBLIC_APP_URL` is unset.** `${process.env.NEXT_PUBLIC_APP_URL ?? ''}/portal/verify?...` — if the env var is missing in any environment (preview, local, mis-provisioned prod) the customer receives `/portal/verify?token=...`, a broken relative URL. Throw on missing instead of silently shipping a broken link.

**10. (LOW) `invoice.total` numeric coercion.** `Number(invoice.total ?? 0)` — if `total` is a Postgres `numeric` returned as string, this works, but if it's `null` due to a draft invoice being marked sent, the customer receives `$0.00`. Add an early validation: a 0-total or null-total invoice should 400, not silently email a $0.00 invoice.

**Verdict: BLOCK MERGE.** Items 1–3 are not negotiable on a critical send-money path. Items 4–6 are exploitable. Items 7–10 should be cleaned up in the same PR to avoid leaving known-broken behavior in a transactional flow.

src/app/api/estimates/[id]/send/route.ts
```diff
@@ -0,0 +1,154 @@
...
+import { sendEstimateReady } from "@/lib/email/sendgrid";
+import { verifyAdminAuth } from "@/lib/firebase/admin-verify";
+import { getServiceRoleClient } from "@/lib/supabase/server-client";
+
+export async function POST(
+  req: NextRequest,
+  { params }: { params: Promise<{ id: string }> }
+): Promise<NextResponse> {
+  try {
+    const admin = await verifyAdminAuth(req);
```
Comment (jack-opsapp) [inline src/app/api/estimates/[id]/send/route.ts:31]: **(BLOCKER) Cross-tenant hole.** `verifyAdminAuth` only validates the JWT signature — it does NOT check that the caller belongs to the same company as the estimate being sent. Any authenticated user in any tenant can POST `/api/estimates/{anyone-elses-id}/send` and trigger a SendGrid email under another company's portal sender. After this check, resolve the caller's `company_id` (via membership lookup) and assert it matches `estimate.company_id` before proceeding. Same issue on the invoice route.

src/app/api/estimates/[id]/send/route.ts
```diff
@@ -0,0 +1,154 @@
...
+      typeof body.email === "string" && body.email.trim().length > 0
+        ? body.email.trim()
+        : null;
+
+    const supabase = getServiceRoleClient();
+
+    // Load estimate
+    const { data: estimate, error: estimateError } = await supabase
+      .from("estimates")
+      .select("id, company_id, client_id, estimate_number, deleted_at")
```
Comment (jack-opsapp) [inline src/app/api/estimates/[id]/send/route.ts:55]: **(BLOCKER) Not idempotent.** This select pulls `status` is missing — and even if it did, there is no `if (estimate.status === 'sent') return 409`. Double-clicking Send (or two tabs, or any retry on a slow network) will fire a second SendGrid email, mint a second portal token, and overwrite `sent_at`. On a money-bearing channel this is a customer-trust failure. Add a `status` to the select and short-circuit with 409 if already sent — let the UI render a `Resend` affirmative-confirm step instead.

src/app/api/estimates/[id]/send/route.ts
```diff
@@ -0,0 +1,154 @@
...
+      );
+    }
+
+    // Load client (for email fallback)
+    const { data: client, error: clientError } = await supabase
+      .from("clients")
+      .select("id, email")
+      .eq("id", estimate.client_id)
+      .single();
+    if (clientError || !client) {
```
Comment (jack-opsapp) [inline src/app/api/estimates/[id]/send/route.ts:74]: **(HIGH) Multi-recipient injection.** This regex does not reject `,`, `;`, `<`, `>`, or whitespace inside the address segments. An operator (or any caller of this route) can submit `a@b.com,c@evil.com` and it passes — then gets handed to `sendEstimateReady` which puts it in SendGrid's `to:` field. Either rejects at SendGrid or splits into multi-recipient send (depending on `gatedSend`). Tighten: reject any address containing `,`, `;`, `<`, `>`, or whitespace; trim; lowercase before passing to BOTH `createPortalToken` and `sendEstimateReady` so the token-bound email matches the recipient header exactly.

src/app/api/estimates/[id]/send/route.ts
```diff
@@ -0,0 +1,154 @@
...
+    const branding = await PortalBrandingService.getBranding(estimate.company_id);
+
+    // Mint a fresh portal token bound to this client + email so the customer
+    // can click straight through to /portal/estimates/{id} without login.
+    const token = await PortalAuthService.createPortalToken(
+      estimate.company_id,
+      estimate.client_id,
+      recipient
+    );
+    const portalUrl = `${process.env.NEXT_PUBLIC_APP_URL ?? ""}/portal/verify?token=${token.token}&redirect=/portal/estimates/${estimate.id}`;
```
Comment (jack-opsapp) [inline src/app/api/estimates/[id]/send/route.ts:110]: **(LOW) Silent broken-link if env missing.** `?? ""` means if `NEXT_PUBLIC_APP_URL` is unset in any environment (preview, mis-provisioned prod, local), the customer receives a relative URL `/portal/verify?token=...` — broken link to the customer's mail client. Throw 500 with a clear error when the env is missing rather than shipping a broken link to a real customer.

src/app/api/estimates/[id]/send/route.ts
```diff
@@ -0,0 +1,154 @@
...
+    const { error: updateError } = await supabase
+      .from("estimates")
+      .update({ status: "sent", sent_at: sentAt })
+      .eq("id", id);
+    if (updateError) {
+      // The email DID send but the DB write failed. Log loudly so ops can
+      // reconcile, and still return 200 — the customer received the doc, and
+      // a manual retry would email them again. UI will refetch and observe
+      // the stale status, but better than double-sending.
+      console.error(
```
Comment (jack-opsapp) [inline src/app/api/estimates/[id]/send/route.ts:137]: **(BLOCKER) Inconsistency window returns lying 200.** When SendGrid succeeds but the DB write fails, the route logs and returns `{success: true, sent_at}`. UI refetches and observes `status` unchanged + `sent_at: null` — operator concludes the send failed and clicks Send again, firing a SECOND email (compounds the missing-idempotency issue above). Either: retry the DB write inline, persist a separate `document_sends` audit row BEFORE the SendGrid call and derive status from it, or return a distinct error code that the UI can render as `email delivered, status stamping failed — do NOT resend`. The current 200-with-success-true is the worst option.

src/app/api/invoices/[id]/send/route.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const branding = await PortalBrandingService.getBranding(invoice.company_id);
+
+    const token = await PortalAuthService.createPortalToken(
+      invoice.company_id,
+      invoice.client_id,
+      recipient
+    );
+    const portalUrl = `${process.env.NEXT_PUBLIC_APP_URL ?? ""}/portal/verify?token=${token.token}&redirect=/portal/invoices/${invoice.id}`;
+
```
Comment (jack-opsapp) [inline src/app/api/invoices/[id]/send/route.ts:110]: **(LOW) Silent $0.00 invoice send.** `Number(invoice.total ?? 0)` coerces null/missing total to 0 and emails the customer an invoice for `$0.00`. A draft-without-total invoice marked sent should 400, not silently mail a zero-dollar invoice. Add an early check: `if (!invoice.total || Number(invoice.total) <= 0) return 400`.

src/components/ops/send-estimate-flow.tsx
```diff
@@ -54,7 +54,7 @@ export function SendEstimateFlow({
 
     setSending(true);
     try {
-      await sendEstimate.mutateAsync(estimate.id);
+      await sendEstimate.mutateAsync({ id: estimate.id, email: email.trim() });
```
jack-opsapp, line 57: **(MEDIUM) Portal-token grant without override confirmation.** The Send dialog accepts any typed email as the recipient (line 107-112) and the backend mints a portal magic-link token bound to whatever the operator typed. There is no second-confirm when `email !== client.email`, and no audit log of the override. A careless or malicious operator can grant `/portal/estimates/{id}` access for this client to an arbitrary address. Add: (a) a confirmation step in the dialog when `email.trim() !== client?.email`, and (b) capture the override in a server-side audit row (`actor_user_id`, `recipient_email`, `client.email`, `portal_token_id`).

jack-opsapp:
Closing — stale (~1 month) and now conflicting with main; the bulk is superseded by intervening work. The still-live fixes this sweep surfaced (critical Stripe default-payment-method; calendar multi-day overlap) are being re-implemented fresh on current main. Reference commits remain available via this PR's refs/pull/54/head.

PR #55, opened 2026-05-11
Nightly Bug Sweep — web — 2026-05-11 — Batch 04

jack-opsapp:
Batch 04 of 04 — 7 bugs (6 fixed, 1 escalated).

## Fixes

- [x] **cc2312ca** (ui) — Collapsed sidebar nav overlapping footer on short viewports. Added `min-h-0` to the nav flex child so `overflow-y-auto` actually clips.  COMPLEX: no.
- [x] **4f5fca8e** (a11y) — Right-edge EdgeTab tabs miss mobile touch + type-size requirements. Tab width now 28px desktop / 44px mobile, wordmark 11px/12px, shortcut chip bumped to 11px universally.  COMPLEX: no.
- [x] **637c9aef** (functional, high) — Mobile sidebar and right-edge drawers can be open simultaneously. DashboardLayout now watches both stores and auto-closes whichever layer was opened first; EdgeTab also returns null on mobile when sidebar is open.  COMPLEX: yes (severity high).
- [x] **8bdcb286** (ui) — Logged-out account-type gate hardcoded copy. Replaced 4 hardcoded strings with `authRequired.*` keys in en + es auth dictionaries.  COMPLEX: no.
- [x] **dfaaa776** (ui) — Login/register use generic SaaS copy. Rewrote 20+ strings across en + es auth dictionaries in OPS command-deck voice. Also routed two hardcoded login error strings (popup blocked, no account) through the dictionary.  COMPLEX: yes (matches keywords auth/login/password/sign in).
- [x] **33e7f258** (a11y) — Password visibility toggles unnamed 14px targets. Both /login and /register toggles now 44×44 hit area, aria-label from dictionary, aria-pressed reflecting state, keyboard-reachable (tabIndex=-1 removed).  COMPLEX: yes (matches keywords auth/login/password).

## Escalated (requires_human_review)

- **f5bd9b27** (ui, high) — Redesigned lockout page overflows on mobile. The bug references `src/components/lockout/` and a worktree path (`.claude/worktrees/lockout-redesign`). Those components do not exist on main; only the old `src/app/(auth)/locked/page.tsx` with an inline `PricingCard`. Per agent instructions, escalating worktree-only.


jack-opsapp:
## Review pass — bug dfaaa776 (login/register OPS auth voice)

**Verdict: STATUS :: REQUEST CHANGES**

**Strong:**
- Voice rewrite is consistent and on-brand: `Resume command`, `Provision account`, `Authenticate via …`, bracketed micro-text `[authenticate to resume operation]`. Drops every "Welcome back", "Get started", "Continue with", "Please …", and "Hold tight".
- Locale parity: `es/auth.json` rewritten in matching voice (`Reanudar mando`, `Provisionar cuenta`, `Autenticar`) — not literal translations of the old SaaS copy.
- Previously hardcoded login error strings (`auth/popup-blocked`, "No account found") now routed through new keys `login.error.popupBlocked` / `login.error.noAccount` in both locales.
- Error specificity preserved: invalid creds → `Invalid email or password`, rate limit → `Too many attempts. Try again later.`, etc.
- Bonus a11y win lands cleanly alongside the dict rewrite: password toggle gets `aria-label` / `aria-pressed` / 44×44 hit target on login. Register page already had the same pattern.
- Headings still proper `<h1>`. Forgot-password / OAuth error paths still pull from dictionary correctly.

**Gap (blocks merge):**
- `src/app/(auth)/register/page.tsx` lines 169-172 still contain two hardcoded English strings using "Please try again" — the exact SaaS phrasing the bug calls out. Spanish users currently see English here. Needs new dictionary keys (e.g. `register.error.syncFailed`, `register.error.syncFailedDetailed`) in both `en/auth.json` and `es/auth.json`, then routed through `t(...)`.

See commit-level comment on `807f78dc` for details.

src/components/layouts/dashboard-layout.tsx
```diff
@@ -191,6 +193,36 @@ export function DashboardLayout({ children }: { children: React.ReactNode }) {
...
+
+    if (sidebarJustOpened && activeEdgeTab !== null) {
+      closeAllEdgeTabs();
+    } else if (edgeJustOpened && sidebarMobileOpen) {
+      closeSidebarMobile();
+    }
+
+    prevSidebarOpen.current = sidebarMobileOpen;
+    prevActiveEdgeTab.current = activeEdgeTab;
+  }, [sidebarMobileOpen, activeEdgeTab, closeAllEdgeTabs, closeSidebarMobile]);
```
jack-opsapp, line 224: REVIEW PASS [bug 637c9aef] :: Effect only fires on a transition (sidebarJustOpened or edgeJustOpened). If a future deep-link or programmatic caller invokes `useEdgeTabStore.setActive()` while the mobile sidebar is already open, neither condition is true — `activeEdgeTab` flips to non-null but the sidebar stays open, leaving the EdgeTab unmounted (via `hiddenForSidebar` in edge-tab.tsx) while the store still reports it active. Result: ghost state. The store's `setActive` is currently only used internally by EdgeTab's onClick, so this is theoretical today, but it is a fragile invariant — anyone who later wires up a deep-link or notification action that programmatically opens a drawer will trip it. Consider also handling the "both already open and activeEdgeTab just changed identity" case, or guarding inside `setActive` itself.

src/components/layouts/dashboard-layout.tsx
```diff
@@ -191,6 +193,36 @@ export function DashboardLayout({ children }: { children: React.ReactNode }) {
...
+  useEffect(() => {
+    const sidebarJustOpened = sidebarMobileOpen && !prevSidebarOpen.current;
+    const edgeJustOpened =
+      activeEdgeTab !== null && prevActiveEdgeTab.current === null;
+
+    if (sidebarJustOpened && activeEdgeTab !== null) {
+      closeAllEdgeTabs();
+    } else if (edgeJustOpened && sidebarMobileOpen) {
+      closeSidebarMobile();
+    }
```
jack-opsapp, line 220: REVIEW PASS [bug 637c9aef] :: Accessibility nit (non-blocking). When the layout auto-closes a drawer here, focus is not restored to the trigger that opened the layer that won (hamburger button or EdgeTab). Screen-reader users get a silent state change. The drawers themselves also lack focus-restore, so this is an existing pattern, not a regression from this commit — but flagging since the fix introduces an *additional* path that closes layers without user action. Future hardening should set focus back to the still-open layer's trigger after the auto-close.

src/components/ui/edge-tab.tsx
```diff
@@ -173,6 +203,10 @@ export function EdgeTab({
   // same min() clamp so the tab + drawer stay flush regardless of viewport.
   const drawerRightOpen = `min(${drawerWidth}px, calc(100vw - ${TAB_WIDTH + 8}px))`;
 
+  if (hiddenForSidebar) {
+    return null;
```
jack-opsapp, line 207: REVIEW PASS [bug 637c9aef] :: Belt-and-suspenders `return null` here unmounts the EdgeTab + its associated drawer (NotificationsDrawer / QuickActionsDrawer / BugReportDrawer all render through `<EdgeTab>`). If the store's `activeTab` is somehow non-null when sidebar opens (e.g. the dashboard-layout effect didn't fire because neither value *transitioned* — see other comment), the drawer disappears but the store still says it's open. Subsequent `useEdgeTabStore.toggle(id)` calls will treat it as already-active and close (correct), but any consumer reading `activeTab` (e.g. analytics, telemetry, sibling-push logic) will see stale state. Today no such consumer exists outside EdgeTab itself, so the fix is safe — flagging for the next person who adds one.

jack-opsapp:
REVIEW PASS [bug 637c9aef] :: Mutual-exclusion fix is sound for all current call paths. `useRef`-tracked transition detection correctly handles same-tick simultaneous opens (sidebar wins deterministically). All three right-edge drawers (notifications / quick actions / bug-report) route through `useEdgeTabStore.activeTab` so the layout effect covers them uniformly. Belt-and-suspenders `hiddenForSidebar` in EdgeTab unmounts the rail entirely on mobile. No double body-scroll-lock concern — none of these drawers lock scroll. Desktop→mobile resize: sidebar's existing resize listener handles the inverse. Z-index split (sidebar z-45 nav band, EdgeTab zIndex 1550 floating-ui band) is fine — they never coexist on mobile after fix.

Three non-blocking flags posted as line comments:
1. Effect only fires on *transitions* — a future programmatic `setActive` while sidebar is already open would leave ghost state. Theoretical today (no such caller).
2. Focus is not restored to the winning layer's trigger after auto-close (a11y).
3. `hiddenForSidebar` unmounts the drawer without clearing store state — same ghost-state risk as #1.

None of these block the fix. The reported overlap on /settings at 390x844 is resolved.

STATUS :: ALLOW MERGE

jack-opsapp:
## Review pass — bug 33e7f258 (password toggle a11y)

Verdict: **STATUS :: REQUEST CHANGES** — the bug as filed (login + register) is fixed, but the same defect pattern still ships unfixed on a third auth route in the same PR scope.

### What the fix does well
- `min-w-[44px] min-h-[44px]` hit area on both /login and /register toggles, with negative margins so the target extends past the field's inner padding without breaking layout. Meets WCAG 2.5.5.
- `aria-label` routed through the auth dictionary (`passwordToggle.show` / `passwordToggle.hide`) and present in both EN and ES — parity confirmed.
- `aria-pressed` reflects state — show vs hide is announced correctly.
- `tabIndex={-1}` removed — toggles are now reachable in tab order, between password field and submit. Natural focus order: label → field → toggle → submit.
- Native `<button type="button">` so Enter and Space both toggle. Browser default focus ring applies.
- A separate password-toggle implementation at `src/app/(auth)/auth/action/PasswordInput.tsx` (used on password-reset action page) already has the 44×44 + aria-label fix — not in this commit but pre-existing.

### Concerns
1. **`/join` has the same bug, unfixed.** `src/app/(auth)/join/page.tsx` lines 514-524 contains an identical password-toggle button inside `Input`'s `suffixIcon` slot with no `aria-label`, no `aria-pressed`, no 44×44 hit area. /join is the team-invite acceptance route — same auth surface, same user-impact class. The bug filing only named /login and /register, but the audit pattern is identical. Recommend: either expand this commit to cover /join, or open a follow-up bug + PR before this batch merges.
2. **Generic `Input` component is the real root cause.** `src/components/ui/input.tsx` lines 61-65 wraps every `suffixIcon` in `<div className="... [&_svg]:w-3.5 [&_svg]:h-3.5">`. Two consequences:
   - Any svg passed via `suffixIcon` is force-resized to 14px regardless of the icon's own className. The Eye/EyeOff icons in /login and /register *look* like 16px in source but render at 14px in the DOM.
   - The wrapper itself does not enforce a minimum touch target — every caller has to manually add 44×44 via negative margins on the inner button. Three callers now do this; the fourth (`/join`) doesn't.
   - Recommend a follow-up to either (a) expose a `suffixAction` prop on `Input` that renders a properly-sized accessible button slot, or (b) drop the `[&_svg]` size override so icons honor their own className.
3. **No visible focus indicator added.** The button relies on browser default focus rings. Given the dark theme and `text-text-3` icon color, the default ring may be low-contrast against the field's `rgba(255,255,255,0.04)` background. Recommend adding `focus-visible:ring-2 focus-visible:ring-ops-accent focus-visible:ring-offset-2 focus-visible:ring-offset-black` to match the pattern used on the primary submit button in `PasswordInput.tsx` (line 73).

### Dictionary parity
EN `passwordToggle.show` / `.hide` and ES `passwordToggle.show` / `.hide` both present in commit 807f78dc.

### Net
The two routes named in the bug filing are correctly fixed. Concern #1 (/join unfixed) and concern #3 (focus ring) should be addressed before merge or as immediate follow-ups; concern #2 is a refactor target.

jack-opsapp:
Closing — stale (~1 month) and now conflicting with main; the bulk is superseded by intervening work. The still-live fixes this sweep surfaced (critical Stripe default-payment-method; calendar multi-day overlap) are being re-implemented fresh on current main. Reference commits remain available via this PR's refs/pull/55/head.

PR #57, opened 2026-05-13
Nightly Bug Sweep — web — 2026-05-13 — Batch 01

jack-opsapp:
## Summary

Seven bugs from the nightly triage batch. Five commits land on `nightly/bugs-web-2026-05-13-batch-01`. All commits pass `npm run type-check && npm run build`.

### `bug-7c90758b` — projects.team_member_ids write path [qa_bugs · high] [COMPLEX]
**Stop writing projects.team_member_ids in ProjectService.update.** The column is server-derived from project_tasks.team_member_ids via a recompute trigger, so any direct write was silently overwritten on the next task edit (users saw project team changes apply momentarily and then revert). Reads remain valid via the trigger-maintained column.

### `bug-a1cc9a86` — Subscription lockout dialog SWITCH ACCOUNT link [bug_reports · medium] [COMPLEX]
**Clear the session before navigating to /login from the lockout dialog.** The previous `<Link href="/login">` was a no-op — middleware saw the still-valid Firebase cookies and bounced /login back to /dashboard, which re-mounted the same lockout dialog. Replaced with an onClick that clears both auth cookies, resets client stores, awaits Firebase signOut(), then runs a full document navigation. Mirrors the cleanup in `src/components/ops/sign-out-overlay.tsx`.

### `bug-9b0f2305` — Project edit workspace missing status + crew controls [qa_bugs · high] [COMPLEX]
**Added a STATUS picker and a TEAM rail to the project edit workspace.** Status lives in the Schedule tab (Archived excluded — it has the footer ARCHIVE action with its own notifications flow). Status changes go through `useUpdateProjectStatus` on submit so the lifecycle hook still posts the activity row and dispatches notifications; only fires when the value actually changed. TEAM section mirrors the viewing sidebar's read-only roster with a callout that team assignments live on tasks — respects bug-7c90758b's constraint that project-level team_member_ids writes are server-overwritten. en + es dictionaries updated.

### `bug-a2dabd69` — Bug report tab not visible on /dashboard [bug_reports · none]
**Removed the /dashboard exclusion in `bug-report-tab.tsx`.** The guard was inherited from the legacy bottom-left floating button era, where the dashboard map filter rail collided with the button. The current edge tab lives on the right rail; that collision rationale no longer applies, and reporters expect to file a bug from the dashboard itself. Kept /intel hidden (intel still owns its own edge controls).

### `bug-ffa94025` — Duplicate `won` opportunities from email sync [bug_reports · urgent]
**Added `won` to the dedup stage window in `/api/integrations/email/import`.** The previous window only matched open pipeline stages, so every inbound for a client whose last lead was already marked won spawned a fresh won row (same client, same job, four rows for the Karen Etheridge sample). `lost` and `discarded` remain excluded as dead-end states. Also added `archived_at IS NULL` so archived rows are not accidentally reused. NOTE: existing duplicate rows in prod still need a one-time cleanup migration, which is out of scope for this batch (migrations owned by another agent).

### `bug-36f8a964` — opportunities.title = "" on email-import [bug_reports · urgent]
**Added `deriveOpportunityTitle` with a four-tier fallback.** The previous write satisfied the NOT NULL constraint with an empty string when the wizard payload's title was null. The new chain: explicit lead.title -> stripped email thread subject -> first sentence of the AI description (cap 80 chars) -> tactical fallback `[OPPORTUNITY · {client} · {YYYY-MM-DD}]`. Always returns a non-empty trimmed string. Existing empty-title rows still need a one-time backfill (out of scope).

### `bug-f64aa932` — clients.phone_number / address sync gap [bug_reports · medium]
**Added `backfillClientContact` to `/api/integrations/email/import`.** Called once per lead after clientId is resolved (covers every branch — merge, link, create_subclient, mergeWithLeadId, create_new). Selects the client row, checks for empty phone_number / address, and updates from lead.clientPhone / lead.clientAddress when present. Fill-blanks only — never overwrites existing values, so the merge-overwrite path remains the canonical edit surface. NOTE: existing client rows that were already created empty still need a one-time backfill from opportunities.contact_phone / projects.address — out of scope.

## Test plan

- [ ] Project workspace editing on a project with a current team — confirm STATUS picker renders with the current value, TEAM section mirrors viewing sidebar roster, callout copy reads correctly
- [ ] Save a project with a changed status — confirm an activity timeline row appears (status_change) and team notifications fire
- [ ] Sign in as a non-admin on a company with `subscription_status = 'lapsed'`, click `// SWITCH ACCOUNT` — confirm the dialog actually unmounts, the cookies are cleared, and /login renders (no redirect back to /dashboard)
- [ ] Visit /dashboard — confirm the bug report tab is now visible on the right rail
- [ ] Run an email import on a client with an existing `won` opportunity — confirm no new row is created and the thread attaches to the existing opp
- [ ] Run an email import on a wizard lead with no explicit title — confirm the resulting opportunity has a non-empty title sourced from the thread subject
- [ ] Run an email import where the lead carries clientPhone / clientAddress and the existing client row is empty — confirm the client row is backfilled


src/lib/api/services/project-service.ts
```diff
@@ -98,7 +98,11 @@ function mapToDb(data: Partial<Project>): Record<string, unknown> {
   if (data.companyId !== undefined) row.company_id = data.companyId;
   if (data.clientId !== undefined) row.client_id = data.clientId;
   if (data.allDay !== undefined) row.all_day = data.allDay;
-  if (data.teamMemberIds !== undefined) row.team_member_ids = data.teamMemberIds;
+  // NOTE: projects.team_member_ids is server-derived from project_tasks.team_member_ids
+  // via a recompute trigger. Writes here are silently overwritten on the next task edit.
+  // Team mutations must go through the assign_project_team_member / remove_project_team_member RPCs,
+  // which mutate task-level assignments. Reads of teamMemberIds remain valid because the column
+  // stays in sync via the trigger. (bug 7c90758b)
```
jack-opsapp, line 105: REVIEW PASS :: BLOCKER

The fix removes the write to projects.team_member_ids inside mapToDb, but every caller still passes teamMemberIds and the createProject path uses this same mapToDb (line 261). Concrete scenario: a user visits /projects/new (src/app/(dashboard)/projects/new/page.tsx:324) and fills in team members. useCreateProject (src/lib/hooks/use-projects.ts:130-150) calls ProjectService.createProject({ teamMemberIds: [...] }), which silently drops the field. Then onSuccess at line 140-148 dispatches a project-assignment push notification to those team members — telling them they have been assigned to a project where they are not actually assigned. The project has no tasks yet, so the recompute trigger never runs, and the team stays permanently empty. Users get a push notification for a non-existent assignment.

Suggested fix: either (a) explicitly throw / strip teamMemberIds inside mapToDb when present so calls that try to write it fail loudly, OR (b) wire the create path to the assign_project_team_member RPC after the insert, OR (c) at minimum, strip teamMemberIds in createProject and useCreateProject and remove the dispatchProjectAssignment call until the RPC path exists.

src/components/lockout/lockout-shell.tsx
```diff
@@ -39,6 +41,37 @@ export function LockoutShell({
...
+      await signOut();
+    } catch {
+      // Best-effort — proceed to /login regardless. The cookies are already
+      // cleared, so middleware will not bounce us back.
+    }
+
+    // Full document navigation to drop any in-memory React state that the
+    // lockout dialog mounted from.
+    window.location.href = "/login";
+  }, []);
```
jack-opsapp, line 74: REVIEW PASS :: CONCERN

The fix swaps a Link for a button that runs an async cleanup and then navigates, but the button never enters a visibly disabled or pending state. The `switchingRef` guard prevents a re-entrant call, but a user who clicks SWITCH ACCOUNT sees nothing change in the UI for the duration of `await signOut()` (typically 100-600ms, but can be slower on flaky networks because Firebase tries to revoke the refresh token). That is the exact failure mode the original bug describes ("the user perceives the button as dead"). A user who clicks twice while `signOut()` is still in flight will hit the early-return on the second call, get no visual feedback, and may conclude the fix did not land.

Compare to `src/components/ops/sign-out-overlay.tsx` — the canonical reference — which mounts a full-screen 4s tactical overlay with three progress bars (CLEARING LOCAL DATA / RELEASING DATABASE CONNECTION / SIGNING OUT [USER]) and only navigates once the animation completes. The lockout SWITCH ACCOUNT click could trigger that same overlay via `useSignOutStore.getState().start({ userName })` instead of duplicating the cleanup inline, getting the visible confirmation for free and matching the rest of the app's sign-out UX.

Suggested fix: replace the inline cleanup with `useSignOutStore.getState().start({ userName: <currentUserName> })` and let `SignOutOverlay` handle cleanup + navigation. If you keep the inline path, at minimum add a pending state (e.g. swap label to `// SIGNING OUT...` and `disabled` the button) so the click is acknowledged immediately.

src/components/lockout/lockout-shell.tsx
```diff
@@ -39,6 +41,37 @@ export function LockoutShell({
...
+    switchingRef.current = true;
+
+    // Clear the two auth cookies up front so the middleware can't read them
+    // on the way to /login.
+    document.cookie = "ops-auth-token=; path=/; max-age=0";
+    document.cookie = "__session=; path=/; max-age=0";
+
+    // Clear client stores so nothing flashes back.
+    useSetupStore.getState().reset();
+    useAuthStore.getState().logout();
```
jack-opsapp, line 62: REVIEW PASS :: NITPICK

TanStack Query caches are not cleared on sign-out here. If the user signs back into a *different* account on the same browser tab (the whole point of SWITCH ACCOUNT), stale query data from the lapsed account stays in memory until the next refetch. Combined with the full-document navigation (`window.location.href = '/login'`), this is actually fine in practice — the page reload drops the entire React tree and re-instantiates the QueryClient — so this is informational only, not a blocker.

The reference pattern in `sign-out-overlay.tsx:51-65` also does not clear the QueryClient, so this fix matches the reference. Worth noting as a known characteristic of the sign-out path rather than a defect introduced by this PR.

src/lib/api/services/project-service.ts
```diff
@@ -98,7 +98,11 @@ function mapToDb(data: Partial<Project>): Record<string, unknown> {
   if (data.companyId !== undefined) row.company_id = data.companyId;
   if (data.clientId !== undefined) row.client_id = data.clientId;
   if (data.allDay !== undefined) row.all_day = data.allDay;
-  if (data.teamMemberIds !== undefined) row.team_member_ids = data.teamMemberIds;
+  // NOTE: projects.team_member_ids is server-derived from project_tasks.team_member_ids
```
jack-opsapp, line 101: REVIEW PASS :: CONCERN

The fix removes the database write but leaves the assignment push-notification dispatch wired to a field that no longer lands. Concrete scenario: any future caller of useUpdateProject that passes a teamMemberIds patch (and the optimistic-update mechanic at line 168-188 absorbs it into the cache) will trigger the cached UI to flip — and a refetch overwrites it. This is the exact user-visible symptom described in bug 7c90758b: 'apply momentarily and then revert.' The fix removes the persisted write but did not remove the optimistic cache mutation, so the bug's UX symptom is still reproducible on the next caller who passes teamMemberIds.

Suggested fix: in useUpdateProject's onMutate, strip teamMemberIds from the optimistic spread (line 182-184), and add an explicit warning/assertion if teamMemberIds shows up in data so future callers can't accidentally trip it.

src/app/api/integrations/email/import/route.ts
```diff
@@ -347,8 +528,10 @@ async function runImport(
           "quoted",
           "follow_up",
           "negotiation",
+          "won",
         ])
         .is("deleted_at", null)
+        .is("archived_at", null)
         .limit(1);
```
jack-opsapp, line 535: REVIEW PASS :: BLOCKER

The dedup window still collapses unrelated jobs from the same client onto a single opportunity row. The key here is purely `(company_id, client_id, stage IN open+won)` with `.limit(1)` and no ordering, no thread_id check, no temporal window. Realistic scenario: Karen Etheridge has a 'Vinyl Railings' opportunity in stage='won' from April. In July she sends a fresh email about a totally separate 'Repair deck stairs' job — the wizard correctly classifies it as a new lead, but at write time this query returns the old won row and the new job silently rides on top of it. Title, description, contact info, the new email thread, all get attached to the wrong opportunity. The activity stream, AI summary, and pipeline UI then conflate two physically distinct jobs.

This is the exact failure mode the original bug report described (4 different jobs collapsed into one client's row history) — just inverted: instead of duplicating, the fix now under-creates and loses real jobs. The reason `lost`/`discarded` were kept out is the same reason `won` arguably should be: a terminal state means the next inbound is a NEW opportunity 90% of the time. Roofing/painting/landscaping clients legitimately come back for repeat business — that's the entire repeat-customer category.

Suggested fix: require an additional match signal before treating an existing won opp as the same job. Options in priority order: (1) Match on `thread_id` via `opportunity_email_threads` — if the incoming `lead.threadId` is already linked to opp X, reuse X regardless of stage. Otherwise create a new opp. This is the actual dedup key. (2) If thread match fails, add a temporal guard — only collapse onto a won opp if its `actual_close_date` is within the last ~14 days (still-in-flight job, not a year-old repeat). (3) Add `.order('created_at', { ascending: false })` so at minimum the newest opp wins instead of arbitrary ordering.

src/components/lockout/lockout-shell.tsx
```diff
@@ -39,6 +41,37 @@ export function LockoutShell({
...
+  // Mirrors the cleanup sequence in src/components/ops/sign-out-overlay.tsx.
+  // (bug a1cc9a86)
+  const handleSwitchAccount = useCallback(async () => {
+    if (switchingRef.current) return;
+    switchingRef.current = true;
+
+    // Clear the two auth cookies up front so the middleware can't read them
+    // on the way to /login.
+    document.cookie = "ops-auth-token=; path=/; max-age=0";
+    document.cookie = "__session=; path=/; max-age=0";
```
jack-opsapp, line 58: REVIEW PASS :: NITPICK

The cookies are SET in `src/components/providers/auth-provider.tsx:21` with attributes `path=/; max-age=2592000; SameSite=Lax`. The clear writes `path=/; max-age=0` — no SameSite. Browsers match cookies for deletion on (name, domain, path) only, so omitting SameSite/Secure on the clear is correct and works. This matches the canonical reference at `sign-out-overlay.tsx:56-57` exactly.

One thing to keep in mind for future hardening: if `__session` were ever set with an explicit Domain attribute (e.g. `.opsapp.co` from a Firebase Hosting rewrite), this clear would silently fail because it has no Domain. Today the app never sets `__session` in client code (confirmed via grep), so this is a non-issue in current state — flagging only as future-proofing context.

src/lib/api/services/project-service.ts
```diff
@@ -98,7 +98,11 @@ function mapToDb(data: Partial<Project>): Record<string, unknown> {
...
+  // via a recompute trigger. Writes here are silently overwritten on the next task edit.
+  // Team mutations must go through the assign_project_team_member / remove_project_team_member RPCs,
```
jack-opsapp, line 103: REVIEW PASS :: BLOCKER

The fix's commit message and the bug-fix notes both say 'Team mutations must now go through the assign_project_team_member / remove_project_team_member RPCs.' Verified by grep across src/: zero callsites for either RPC name exist in the web codebase. The fix is deletion-only — it removes the write path without supplying the replacement path. Concrete consequence: there is currently no working way for a user on web to change a project's team via the project surfaces. The workspace-edit form does not pass teamMemberIds (so that surface is silent and the user is confused), and /projects/new and useProjectMutations.createProject both pass teamMemberIds which is silently dropped. The bug's promised remediation ('Team mutations must go through the RPCs') is not implemented anywhere.

Suggested fix: add a thin service method on ProjectService — e.g. assignProjectTeamMember(projectId, userId) and removeProjectTeamMember(projectId, userId) — that calls supabase.rpc('assign_project_team_member', ...) and wire useProjectMutations.saveProject / useCreateProject to call these for the added/removed deltas after the row update lands.

src/lib/api/services/project-service.ts
```diff
@@ -98,7 +98,11 @@ function mapToDb(data: Partial<Project>): Record<string, unknown> {
...
+  // which mutate task-level assignments. Reads of teamMemberIds remain valid because the column
```
jack-opsapp, line 104: REVIEW PASS :: CONCERN

The mapToDb function silently drops the teamMemberIds field instead of failing loudly. With 19+ files in the codebase that reference teamMemberIds (per grep), and Partial<Project> as the input type, any future engineer writing code like 'ProjectService.updateProject(id, { teamMemberIds: [...] })' will produce a no-op with no compile-time, runtime, or log error — the exact silent-failure pattern that caused this bug in the first place.

Suggested fix: add an explicit guard at the top of mapToDb, e.g.:
  if (data.teamMemberIds !== undefined) {
    console.warn('[project-service] teamMemberIds is server-derived; use assignProjectTeamMember RPC instead. Ignoring.');
  }
or strengthen the type signature of updateProject/createProject to omit teamMemberIds at the type level (Omit<Partial<Project>, 'teamMemberIds'>) so the silent drop becomes a compile-time error.

jack-opsapp:
REVIEW PASS (bug a1cc9a86) :: 3 concerns (0 blockers, 1 concern, 2 nitpicks)

STATUS :: ALLOW MERGE

The fix correctly clears both auth cookies BEFORE awaiting Firebase signOut() and uses a full document navigation — the cookie clear ordering means even if signOut() throws, middleware will not bounce the user back to /dashboard. The clear pattern is byte-identical to the canonical reference in sign-out-overlay.tsx:56-57. switchingRef guards re-entrancy. Auth state ordering, cookie attributes, and error paths all check out.

Concerns posted:
- CONCERN: No pending state on the button — user clicking SWITCH ACCOUNT sees no visual feedback during await signOut(), which is the exact perception failure mode the original bug describes. Recommend triggering the existing SignOutOverlay via useSignOutStore instead of duplicating cleanup inline.
- NITPICK: TanStack Query cache not cleared (matches reference pattern; covered by full-page reload).
- NITPICK: Cookie clear has no Domain attribute — fine today since __session is never set client-side with a Domain, future-proofing note only.

jack-opsapp:
REVIEW PASS (bug 7c90758b) :: 4 concerns (2 blockers, 2 concerns, 0 nitpicks)

STATUS :: BLOCK MERGE

Summary of blockers:
- The fix is deletion-only. The promised replacement path (assign_project_team_member / remove_project_team_member RPCs) is not implemented anywhere in src/ — grep returns zero callsites. There is currently no working way for a user to change a project's team via web project surfaces.
- The create path has the same bug as the update path now. /projects/new and useProjectMutations.createProject both pass teamMemberIds, mapToDb silently drops it, and the post-create dispatchProjectAssignment still fires push notifications to team members who were never actually assigned. The recompute trigger does not fire on project insert (no tasks exist yet), so the team stays permanently empty while the user thinks they assigned a team.

Additional concerns:
- useUpdateProject (use-projects.ts:168-188) still does an optimistic cache update on the full patch, including teamMemberIds. The 'apply momentarily then revert' UX described in the bug is still reproducible by any caller that passes teamMemberIds because the cache flips before the refetch overwrites.
- mapToDb silently drops teamMemberIds instead of failing loudly — same silent-failure shape that caused this bug. Recommend a console.warn or a tighter type on updateProject/createProject (Omit<Partial<Project>, 'teamMemberIds'>) to surface bad calls.

src/app/api/integrations/email/import/route.ts
```diff
@@ -347,8 +528,10 @@ async function runImport(
           "quoted",
           "follow_up",
           "negotiation",
+          "won",
         ])
         .is("deleted_at", null)
+        .is("archived_at", null)
```
jack-opsapp, line 534: REVIEW PASS :: CONCERN

When dedup finds an existing opportunity, the code reuses `opportunityId` and skips the entire create/update block — it never refreshes stage, title, description, contact info, last_inbound_at, correspondence_count, or ai_summary on the existing row. Concrete scenario: an opp sits at `qualifying` from last week's email. New inbound arrives today and the wizard correctly classifies it as `negotiation` (operator quoted in the wizard step). The route reuses the existing opp at `qualifying` and silently discards the new stage. Same for: a clearer subject line from the latest email never overwrites the empty title on an old row (so the bug-36f8a964 title fix has no effect on already-existing empty-title rows that get reused), a new phone number in the latest email signature never reaches `contact_phone`, and `last_inbound_at` stays stale so the pipeline-aging UI shows the wrong age.

Suggested fix: after picking the existing opportunityId, apply a targeted patch — `stage` (if the new stage is further along the pipeline), `title` (if existing is empty), `description`/`ai_summary` (append or replace), `correspondence_count`, `inbound_count`, `last_inbound_at`/`last_outbound_at`, `contact_phone` (fill-blanks like the client backfill). Stage transitions should be one-way: never demote (e.g. don't drop a won opp back to qualifying because a confirmation reply came in).

src/components/ops/projects/workspace/edit-create/project-edit-create-body.tsx
```diff
@@ -207,6 +228,15 @@ export function ProjectEditCreateBody({
...
       visibility: project.visibility ?? "all",
+      // Archived projects shouldn't normally route through the editor at all,
+      // but if they do we fall back to RFQ in the picker (the schema doesn't
+      // accept Archived as a writable option). The display title bar still
+      // reads the actual status from the loaded project.
+      status: STATUS_VALUES.includes(
+        project.status as (typeof STATUS_VALUES)[number],
+      )
+        ? (project.status as (typeof STATUS_VALUES)[number])
+        : ProjectStatus.RFQ,
```
jack-opsapp, line 239: REVIEW PASS :: BLOCKER

Archived projects silently un-archive on save. Scenario: dispatcher opens an archived project in the editor (deep link or stale tab), the picker falls back to RFQ here because Archived is not in STATUS_VALUES. They tweak the title/address and submit. `values.status === 'RFQ'` and `project.status === 'Archived'` — the diff check at line 308 is true, so `updateStatus.mutateAsync({ status: 'RFQ' })` fires. The lifecycle handler writes a status_change activity row, dispatches a notification to the crew, and the project rejoins the active pipeline. There's no operator intent here — the picker just defaulted.

Suggested fix: when the loaded project's status is Archived, either (a) suppress the status row entirely in the schedule tab and keep the form value sentinel `Archived` so the diff check evaluates false, or (b) render the picker disabled with the archived hex dot + a hint pointing at the footer's UNARCHIVE action. The status diff guard at line 308 must also explicitly skip when `project.status === ProjectStatus.Archived` regardless of `values.status`.

src/app/api/integrations/email/import/route.ts
```diff
@@ -360,10 +543,17 @@ async function runImport(
...
+          threadId: lead.threadId,
+          description: lead.description,
+          clientName: lead.clientName,
+        });
+
         const opp = await OpportunityService.createOpportunity({
           companyId,
           clientId,
-          title: lead.title || "",
+          title: derivedTitle,
```
jack-opsapp, line 556: REVIEW PASS :: CONCERN

The original bug describes TWO problems for Karen Etheridge: (a) four duplicate `won` rows, and (b) none of them linked to a `project_id`. The fix addresses (a) going forward, but (b) is still un-wired: `projectId` is passed as `null` to createOpportunity below, and there's no other write here that backfills it when `stage === Won`. After this PR ships, an operator confirming a won lead in the inbox wizard will still produce an opportunity with no project link — and the WORK tab will keep dropping that row out (which the tactical filter in commit 67fc6dcd masks).

If the won→project transition lives elsewhere (likely a separate 'convert to project' action surface), scope this back to the bug ticket: who is supposed to wire `opportunities.project_id` when an inbox-wizard lead is auto-classified as won, and where does that handoff happen? At minimum, the fix should either (i) call into the won-conversion service from this route when `stage === Won`, or (ii) document explicitly in the fix_notes that the won→project link is a separate ticket so the cleanup-migration agent backfills both columns when it lands.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+      .eq("id", args.clientId)
+      .maybeSingle();
+    if (!row) return;
+
+    const update: Record<string, string> = {};
+    const currentPhone = ((row.phone_number as string | null) ?? "").trim();
+    const currentAddress = ((row.address as string | null) ?? "").trim();
+
+    if (!currentPhone && args.leadPhone && args.leadPhone.trim()) {
+      update.phone_number = args.leadPhone.trim();
```
jack-opsapp, line 162: REVIEW PASS :: CONCERN

Phone numbers arrive in many shapes from AI extraction ('(250) 538-8994', '+12505388994', '250-538-8994', etc.). This backfill stores the raw .trim()'d value, so two leads for the same client with different formats will look 'different' to any future de-dupe or string-equality check elsewhere in the app. Concrete scenario: client gets backfilled with '(250) 538-8994' from a quote form, later a merge with mergeMode='overwrite' from a different thread writes '+12505388994' as phone_number — operators see what looks like the phone 'changing' even though it's the same number, and downstream search/match-by-phone breaks.

Suggested fix: normalize to E.164 (or at least digits-only with a leading + if international) before writing — both here AND in the merge branch (line ~393, ~398) so the canonical phone shape is identical across every write path.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+  if (args.threadId) {
+    try {
+      const { data: thread } = await supabase
+        .from("email_threads")
+        .select("subject")
+        .eq("id", args.threadId)
+        .maybeSingle();
+      const subject = ((thread?.subject as string) ?? "").trim();
+      const cleaned = stripSubjectPrefixes(subject);
+      if (cleaned) return cleaned;
```
jack-opsapp, line 106: REVIEW PASS :: CONCERN

Tier 2 (email thread subject) has no length cap, while Tier 3 (AI description) caps at 80 chars. Real subjects can be 200+ characters (long auto-generated subjects, mailing list prefixes, multi-language). The DB column is unbounded text so no constraint violation, but lead-card UI slots assume a reasonable title length — a 300-char subject will overflow / truncate awkwardly and the operator never sees a clean title.

Suggested fix: after stripSubjectPrefixes returns, apply the same 80-char cap as firstSentence (slice + ellipsis if longer). Keeps the two tiers visually consistent in the UI.

src/app/api/integrations/email/import/route.ts
```diff
@@ -347,8 +528,10 @@ async function runImport(
           "quoted",
           "follow_up",
           "negotiation",
+          "won",
```
jack-opsapp, line 531: REVIEW PASS :: NITPICK

Race condition window: the dedup lookup → createOpportunity path is application-level only. There's no DB unique constraint on `(company_id, client_id)` for opportunities (and there shouldn't be — a client can legitimately have multiple opps). If two webhook deliveries for the same client land within the same few hundred milliseconds (e.g. a Gmail batch redelivery, two inbox-wizard runs from two operator tabs, or one cron firing while another is mid-flight), both queries return zero existing opps and both insert. The race window is small but real, and this route's behavior under contention is what produced the duplicate-row outcome in the first place.

Not a blocker for this batch — but worth a follow-up: either serialize import jobs per (company_id, client_id, thread_id) at the route level, or add a partial unique index on `(company_id, client_id, thread_id_of_first_attached_email)` — whichever maps cleaner to your existing patterns.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+      update.phone_number = args.leadPhone.trim();
+    }
+    if (!currentAddress && args.leadAddress && args.leadAddress.trim()) {
+      update.address = args.leadAddress.trim();
+    }
+    if (Object.keys(update).length === 0) return;
+
+    const { error } = await supabase
+      .from("clients")
+      .update(update)
```
jack-opsapp, line 171: REVIEW PASS :: CONCERN

The update at line 171 has no company_id filter. The clients table is multi-tenant and this code runs under the service-role key, which bypasses the database-level row filtering (RLS). In practice the clientId here is always derived from this company's flow (the create-new branch filters by company_id, the existingClientId comes from a wizard payload that's company-scoped), so a cross-tenant write is not currently reachable. But this is one stack-trace away from being a real cross-tenant write — e.g. a future bug where existingClientId arrives from a less-trusted source, or a refactor that detaches clientId from the company flow. Defense-in-depth: add .eq('company_id', companyId) on the update query (and on the SELECT at line 153) so a stray clientId from a different tenant can never resolve. The merge branch (line 405) has the same gap.

src/components/ops/projects/workspace/edit-create/project-edit-create-body.tsx
```diff
@@ -271,6 +301,16 @@ export function ProjectEditCreateBody({
...
+      // Status flips go through the lifecycle hook so the activity timeline +
+      // notifications fire identically to every other surface that updates
+      // status. Only call it when the operator actually changed the value to
+      // avoid spurious status_change events / notifications. (bug 9b0f2305)
+      if (project && values.status !== project.status) {
+        await updateStatus.mutateAsync({
+          id: projectId,
+          status: values.status,
+        });
+      }
```
jack-opsapp, line 313: REVIEW PASS :: CONCERN

Two sequential mutations create a visible cache flicker when both run. `saveProject` has no `onMutate` — it relies on `onSuccess` → `invalidateQueries` to refresh the cache. Meanwhile, immediately after `saveProject.mutateAsync` resolves, `updateStatus.mutateAsync` runs and its own `onMutate` (use-projects.ts line 263) reads the cached `previousProject` and spreads it back with the new status: `{ ...previousProject, status }`. If the invalidation refetch hasn't completed yet, `previousProject` still holds the pre-save snapshot (old title/address/etc.). The optimistic write at use-projects.ts line 273 will overwrite the freshly saved fields with their stale values until the second invalidation lands. Title bar, schedule rail, every workspace consumer flickers back briefly.

Worse: if `updateStatus` then fails (network blip), `onError` (line 282) restores `previousProject` permanently — leaving the cache showing the pre-save title even though the database has the new title. The user sees the save get rolled back when only the status update actually failed.

Suggested fix: change the save flow to a single `updateProject` call that includes status, and route the lifecycle plumbing inside `ProjectService.updateProject` when the status field is present (or expose a combined service method). Alternatively, make `saveProject` await an explicit refetch of `projects.detail` before `updateStatus` runs, so the optimistic snapshot is fresh.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+
+  // Tier 2 — email thread subject. Best-effort: a missing row or a
+  // permission failure falls through silently.
+  if (args.threadId) {
+    try {
+      const { data: thread } = await supabase
+        .from("email_threads")
+        .select("subject")
+        .eq("id", args.threadId)
+        .maybeSingle();
```
jack-opsapp, line 103: REVIEW PASS :: CONCERN

This email_threads lookup runs on the service-role supabase client (line 197 of POST handler), which bypasses tenant isolation rules entirely. The query filters only on thread id — no .eq("company_id", companyId). If a wizard payload ever carried a threadId from a different tenant (e.g. malformed client, replay, future code path that crosses tenants), this would silently pull THAT tenant's subject and write it as the title of THIS tenant's opportunity row — a quiet cross-tenant data leak.

Scenario: operator A's wizard somehow receives a thread id that happens to collide / be predicted / be malicious-input for operator B's thread. Service-role lookup succeeds, B's subject ends up in A's opportunity.title, visible across A's UI.

Suggested fix: add .eq("company_id", companyId) to the email_threads query. Threading companyId into deriveOpportunityTitle takes one extra arg. Defense-in-depth — even if the threat surface is small today, this is a free constraint.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+  const match = trimmed.match(/^[^.!?\n]+/);
+  const candidate = (match ? match[0] : trimmed).trim();
+  if (!candidate) return null;
+  if (candidate.length <= max) return candidate;
+  return candidate.slice(0, max - 1).trimEnd() + "…";
+}
+
+function tacticalFallbackTitle(clientName: string | null | undefined): string {
+  const safeName = (clientName ?? "").trim() || "UNKNOWN CLIENT";
+  const today = new Date().toISOString().slice(0, 10);
```
jack-opsapp, line 78: REVIEW PASS :: NITPICK

tacticalFallbackTitle uses new Date().toISOString().slice(0, 10) — that's the UTC date, not the operator's local date. For a Vancouver operator running an import at 8pm PT, the import generates titles dated tomorrow (UTC has already rolled over). It's the tier-4 last-resort fallback so this rarely shows in practice, but when it does it looks like a wrong-date bug to the operator.

Suggested fix: use a timezone-aware format (e.g. Intl.DateTimeFormat with the company's timezone, or simply the operator's local date if available on the import payload). Or accept this as a known quirk since the fallback should be rare.

src/components/ops/projects/workspace/edit-create/schedule-tab.tsx
```diff
@@ -167,8 +272,79 @@ export function ScheduleTab() {
...
+        ) : (
+          <ul
+            data-testid="schedule-team-list"
+            className="flex flex-col gap-1"
+          >
+            {members.map((m) => (
+              <li
+                key={m.id}
+                className="flex items-baseline justify-between gap-2"
+              >
```
jack-opsapp, line 320: REVIEW PASS :: CONCERN

The TEAM rail shows 'No one assigned' before upstream queries resolve. `useProjectTeam` returns a derived `members` array but no loading flag — when `useTeamMembers()`, `useProjectTasks()`, or `useTaskTypes()` are still in flight, `members` is `[]` and the empty-state copy 'No one assigned.' renders. Scenario: dispatcher opens a project that has six crew members assigned via tasks. For a brief moment they see the empty state copy and might think the team data was wiped. Then it pops in.

Suggested fix: have `useProjectTeam` return `{ members, isLoading }` by aggregating `isLoading` from the four upstream hooks, and gate the empty state branch on `!isLoading` in TeamSummarySection. While loading, render a skeleton (or just nothing — anything but the false-negative empty state).

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+        `[email-import] Client contact backfill failed for ${args.clientId}: ${error.message}`,
+      );
+    }
+  } catch (err) {
+    console.error(
+      `[email-import] Client contact backfill threw for ${args.clientId}:`,
+      err,
+    );
+  }
+}
```
jack-opsapp, line 184: REVIEW PASS :: CONCERN

The original bug report explicitly calls out projects.address -> clients.address as part of the same sync gap: 'Same for projects.address -> clients.address when a project's address is set and client's is not.' The current fix only addresses lead.clientAddress -> clients.address at email-import time. If a project gets created (or has its address edited) and the client row is still empty, the client row stays empty. The 25% address coverage stat from the bug includes that path.

Concrete scenario: a 'won' opportunity converts to a project with an operator-typed address; the original client was created from a sparse Gmail thread with no address; the client surface still shows blank address even though the project has it.

Suggested fix: add a sibling backfill in the project create / update path that fills client.address from project.address when the client value is empty. Can be a follow-up PR if scope is tight here, but it should be tracked — please file a follow-up and link it in fix_notes.

jack-opsapp:
REVIEW PASS (bug ffa94025) :: 4 concerns (1 blocker, 2 concerns, 1 nitpick)

STATUS :: BLOCK MERGE

Summary of blockers:
- The dedup key is purely (company_id, client_id, stage IN open+won) with limit(1) and no thread_id / temporal / ordering signal. Including 'won' in the in-list now causes the OPPOSITE bug: a client's brand-new unrelated job will silently collapse onto their year-old won opportunity. Repeat-customer trades (roofing, painting, landscaping) will see real jobs disappear into the wrong row. Needs thread_id-based matching or a temporal guard before 'won' can safely be in the dedup window. See line 535 comment for suggested fix.

Other concerns:
- Existing opportunity branch never updates stage / title / contact / last_inbound_at on reused rows — operator-confirmed wizard changes get silently dropped (line 534).
- Original bug had two halves: duplicate wons AND missing project_id. This PR only addresses the first half. project_id is still hardcoded null at write time, so the WORK-tab phantom-row issue persists even after this ships (line 556).
- Race condition between concurrent webhook deliveries for the same client — application-level lookup-then-insert with no DB unique constraint. Worth a follow-up ticket (line 531).

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+//      Looked up directly from the email_threads row by thread id.
+//   3. The first sentence (up to 80 chars) of the AI-generated description.
+//   4. A tactical fallback: `[OPPORTUNITY · {client} · {YYYY-MM-DD}]`.
+//
+// Steps 1 and 4 always work synchronously. Steps 2 and 3 are best-effort —
+// failures fall through to the next tier. Result is always a non-empty
+// trimmed string, which is what the NOT NULL constraint actually wants.
+// (bug 36f8a964)
+
+const SUBJECT_PREFIX_RE = /^\s*(re|fwd?|fw):\s*/i;
```
jack-opsapp, line 52: REVIEW PASS :: NITPICK

SUBJECT_PREFIX_RE only matches English prefixes (re, fwd, fw). Real inboxes pick up multilingual replies — "AW:" (German Antwort), "WG:" (German Weitergeleitet), "RV:"/"REF:" (Spanish), "TR:" (French Transféré), "Antw:" (Dutch), "OD:" (Polish). For a client like Karen Etheridge today, irrelevant. For a contractor doing cross-border work or any forwarded chain from a European supplier, the Re/Fwd stack persists as visible prefix noise on the lead title.

Suggested fix (low priority): broaden the regex to include common locale prefixes, e.g. /^\s*(re|fwd?|fw|aw|wg|rv|ref|tr|antw|od):\s*/i. Or accept the English-first MVP scope.

Also: if the subject contains embedded newlines (rare, but RFC 5322 allows folding), stripSubjectPrefixes doesn't collapse them. firstSentence handles newlines via the regex; Tier 2 does not. A subject like "Re: roof\n\nplease quote" would write the literal newline to opportunities.title.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+
+  // Tier 3 — first sentence of the AI description.
+  if (args.description) {
+    const sentence = firstSentence(args.description, 80);
+    if (sentence) return sentence;
+  }
+
+  // Tier 4 — deterministic fallback in OPS tactical voice.
+  return tacticalFallbackTitle(args.clientName);
+}
```
jack-opsapp, line 123: REVIEW PASS :: CONCERN

No regression test was added for deriveOpportunityTitle even though it's the entire fix surface for an urgent bug. The four-tier fallback chain has at least eight branching cases (explicit-trim-empty, explicit-present, thread-missing-row, thread-empty-subject, thread-stack-of-prefixes, description-only, description-empty, fallback-with-null-clientName). A unit test file at tests/unit/email-import/title-derivation.test.ts covering each tier (with the supabase client mocked for tier 2) would lock in the contract: future refactors can't quietly slide back to title="".

Out-of-scope is fine for this PR if the fix is intentionally minimal, but worth tracking as a follow-up because the bug originated precisely from an untested write path.

src/components/ops/projects/workspace/edit-create/schedule-tab.tsx
```diff
@@ -83,10 +149,49 @@ export function ScheduleTab() {
...
+                  />
+                  <div className="flex-1">
+                    <Select
+                      options={statusOptions}
+                      value={field.value}
+                      onChange={(v) =>
+                        field.onChange(v as ProjectEditCreateFormValues["status"])
+                      }
+                      placeholder={t("schedule.status.placeholder")}
+                    />
```
jack-opsapp, line 188: REVIEW PASS :: NITPICK

The STATUS Field label is not programmatically tied to the Radix Select trigger. The Field atom renders the label, but the Controller injects a div containing a swatch span and the Select — and no `id` flows from Field through to the Select trigger. Screen readers will read the label text on focus only via DOM proximity, not via `aria-labelledby` / `htmlFor`. Keyboard navigation still works (Radix handles arrow keys and Space), so it's not a blocker, but it deviates from how the Visibility Segmented control or other workspace fields wire up.

Suggested fix: have Field generate a stable id, forward it into Select via `id={`...`}`, and have the Select component pass it to the SelectPrimitive.Trigger.

src/app/api/integrations/email/import/route.ts
```diff
@@ -319,6 +473,18 @@ async function runImport(
...
+      // payload when the client row is empty. Covers every branch above
+      // (merge, link, create_subclient, mergeWithLeadId, create_new)
+      // because they all converge on a single clientId at this point.
+      // Fill-blanks only — never overwrites an existing client value.
+      // (bug f64aa932)
+      await backfillClientContact(supabase, {
+        clientId,
+        leadPhone: lead.clientPhone,
+        leadAddress: lead.clientAddress,
+      });
```
jack-opsapp, line 486: REVIEW PASS :: CONCERN

The fix is forward-only. The bug stats show 75/172 (44%) phone coverage on clients vs 89/172 (52%) on opportunities, and 43/172 (25%) address coverage on clients vs 65/172 (38%) on opps/projects — meaning 14 clients have phone in an opp but not on the client row, and 22 clients have address downstream but not on the client row. The fix only catches new imports going forward; the existing gap stays.

Fix notes call this out as out-of-scope ('existing client rows that were already created empty still need a one-time backfill from opportunities.contact_phone / projects.address — out of scope (no migrations)'), but for an operator, the experience is still 'I see blank client phone numbers' until those clients get re-imported. The web inbox now does the read-time fallback through opp/project (mentioned in the bug report), so the operator-visible value isn't blank — but the underlying clients row is still wrong, which means any other surface that reads clients.phone_number directly (and doesn't chain) will still show blank.

Suggested follow-up: ship a one-time data-fix script (admin-only endpoint or SQL) that walks clients WHERE phone_number IS NULL AND EXISTS opp.contact_phone IS NOT NULL and copies the value over. Track in fix_notes.

jack-opsapp:
REVIEW PASS (bug 9b0f2305) :: 4 concerns (1 blocker, 2 concerns, 1 nitpick)

STATUS :: BLOCK MERGE

Summary of blockers:
- Archived projects silently un-archive on save. The picker defaults to RFQ when an archived project loads (Archived is not in STATUS_VALUES). `values.status !== project.status` evaluates true even though the operator never touched the picker, so `updateStatus.mutateAsync({ status: 'RFQ' })` fires — writing a status_change activity row, notifying the crew, and resurrecting the project into the active pipeline.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+
+    const { error } = await supabase
+      .from("clients")
+      .update(update)
+      .eq("id", args.clientId);
+    if (error) {
+      console.error(
+        `[email-import] Client contact backfill failed for ${args.clientId}: ${error.message}`,
+      );
+    }
```
jack-opsapp, line 177: REVIEW PASS :: NITPICK

No audit breadcrumb when this fires. If an operator opens a client tomorrow and sees a phone they didn't type, there's no way to trace it back to which lead import filled it in. The project workspace uses project_notes with event_kind for exactly this kind of system-event narrative — clients don't have an equivalent table today, but at minimum a console log with both clientId and the lead.id / threadId source would help when investigating future complaints (e.g. 'wrong phone showed up on my client'). The current error path logs on failure; success is silent.

Suggestion: log at info level on success — '[email-import] Client contact backfilled: client={clientId} phone={was filled? yes/no} address={was filled? yes/no} source-lead={lead.id} source-thread={lead.threadId}'.

jack-opsapp:
REVIEW PASS (bug 36f8a964) :: 5 concerns (0 blockers, 3 concerns, 2 nitpicks)

STATUS :: ALLOW MERGE

The four-tier title derivation correctly closes the title='' write-path bug. Tier 1 (explicit), Tier 2 (cleaned thread subject), Tier 3 (first sentence of description), Tier 4 (tactical bracket fallback) collectively guarantee a non-empty trimmed string under every input shape I could construct. No blocker found.

Concerns to consider for follow-up (none gate this merge):
- Tier 2 lacks the same 80-char cap as Tier 3 — long subjects will produce ugly title rows in the UI.
- The email_threads lookup runs under service-role and omits .eq('company_id', companyId) — defense-in-depth gap, not actively exploitable today.
- No regression test was added for deriveOpportunityTitle — the original bug came from an untested write path, so locking the contract is worth a follow-up.

Nitpicks:
- tacticalFallbackTitle uses UTC date instead of operator-local date.
- SUBJECT_PREFIX_RE is English-only; AW:/WG:/RV:/TR: stay stacked on multilingual threads. Embedded newlines in subject not collapsed.

Also noted: the existing rows with title='' still need a one-time backfill — explicitly out-of-scope per the commit message, but the operator-facing symptom for the Karen Etheridge rows won't disappear until that runs.

src/app/api/integrations/email/import/route.ts
```diff
@@ -29,6 +29,160 @@ import type { SupabaseClient } from "@supabase/supabase-js";
...
+// already-existing client row). Operators saw blank contact details on the
+// client surface even though the lead they were looking at had a phone
+// number. (bug f64aa932)
+//
+// On every confirmed lead, we now check whether the linked client row is
+// missing phone or address and the lead payload carries those values — if
+// so we update the client row in place. This is fill-blanks only; we never
+// overwrite an existing client phone/address with import payload values
+// (the merge flow above is the canonical "overwrite" path).
+async function backfillClientContact(
```
jack-opsapp, line 140: REVIEW PASS :: NITPICK

No regression test was added with this fix. backfillClientContact has clear branching behavior (skip-when-both-null, skip-when-existing, fill-only-empty, error path) and is a textbook unit-test target. A small test (mock supabase, assert update payload shape across each branch) would lock this down against future refactors of the import path. Same for deriveOpportunityTitle — its 4 tiers are deterministic and easy to cover.

Not blocking, but worth filing as a follow-up.

jack-opsapp:
REVIEW PASS (bug f64aa932) :: 5 concerns (0 blockers, 3 concerns, 2 nitpicks)

STATUS :: ALLOW MERGE

The fill-blanks backfill is correct for the forward-going path and the fix notes accurately scope what's in and out. No blockers — the fix is safe (never overwrites existing data, supabase is service-role so the read+update isn't blocked by row-level security, the clientId is always company-scoped through the surrounding flow).

Concerns to address (follow-up, not blocking):
- Phone-format normalization (line 162): no E.164 canonicalization, two formats of the same number will round-trip as 'different'.
- Cross-tenant defense in depth (line 171): update has no company_id filter — currently safe because clientId is always company-scoped, but one refactor away from a real cross-tenant write.
- projects.address -> clients.address gap (line 184): bug explicitly calls this out; this PR only covers lead.clientAddress -> clients.address.

Nitpicks:
- No audit / breadcrumb log on success.
- No regression test added.

jack-opsapp:
Closing — stale (~1 month) and now conflicting with main; the bulk is superseded by intervening work. The still-live fixes this sweep surfaced (critical Stripe default-payment-method; calendar multi-day overlap) are being re-implemented fresh on current main. Reference commits remain available via this PR's refs/pull/57/head.

